Issue #31, opened 2017-02-14
Improve the "Loading" screen


EricWittmann:
The loading screen is what the user sees when the application is initially loading.  The markup for this is found in the host/index page and should be improved (currently it's just "Loading..." text!).

msavy:
I've been playing around and (trying to _start_ getting up-to-date with the new CSS animation stuff). Hope I might be able to contribute something fun!

Here's a first little prototype idea, hopefully do something fancier shortly.

The idea is that apiman fades in on load, then fades out as/when load completes (or similar). It uses pure HTML & CSS but I'd imagine a bit of JS might be necessary for real the implementation. 

![image](https://i.imgur.com/QyMGIL2.gifv)

msavy:
Here's a codepen (animations seem smoother in chrome FWIW). 

https://codepen.io/marcsavy/pen/WRmMmy

EricWittmann:
Oh that's very nice!  PR forthcoming?  :)

msavy:
I'd be very glad to if you think it looks good enough!

msavy:
BTW, I hacked the SVG in Inscape for that presentation I did at JBCNConf, and I'm no pro with it, so it's probably slightly wonky - hah.

msavy:
How's about this one for April 1st ;-) 

https://codepen.io/marcsavy/pen/MJxXYQ

msavy:
I'm nearly ready to make a PR for you to preview. One small thing, which do you prefer:

Unmodified:
<img src="https://cloud.githubusercontent.com/assets/423513/22996441/b19f3716-f3c6-11e6-98d0-38925da76cc4.png" width=50%>

With `border-radius: 10%`:
<img src="https://cloud.githubusercontent.com/assets/423513/22996406/8d20e40c-f3c6-11e6-9fb7-94f052be636e.png" width=50%>

More extreme (50%)

<img src="https://cloud.githubusercontent.com/assets/423513/22996835/0f39fa2c-f3c8-11e6-9e41-a167ac54dbb7.png" width=50%>




Here's a codepen: NB: I set a darker background to make it easier to design, but in reality it's a lighter colour to match the normal UI background colour as you can see in the pics. Any weirdness with corner rounding can be fixed relatively easily, I think - just gauging what you think. 

https://codepen.io/marcsavy/pen/ggEEoK

EricWittmann:
I like the 10% one.  Looks nice!

msavy:
[...]

PR #40, opened 2017-02-17
Lines for hero to increase contrast.

msavy:
I tried alternatives but it seems you can't do SVG manipulations unless its inline, so I added hero directly into index.html.

If you feel it's a problem I can try to find an alternative.


EricWittmann:
Nice, thanks!  Still fairly subtle but I actually think that's a good thing. :)

msavy:
I have definitely been aiming for subtle - but just am aware that different displays can make it appear quite different!

msavy:
Resolves #31

PR #88, opened 2017-03-28
Adding Docker Generation Files

maneta:
This Adds a Dockerfile for container image generation and a Makefile to operate the image. ```make help``` provides help for the Makefile targets.

Please feel free to change the docker registry namespace. I was using my personal namespace (maneta) for  to test the deployment. 

The requirements to run this are:

* GNU Make
* Docker

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+#Avoid large conext send to Docker Deamon
+*
+!front-end/quickstart/target/api-design-studio-*-quickstart.zip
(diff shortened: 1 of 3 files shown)


EricWittmann:
Thanks!  I'll give this a try tomorrow.

maneta:
Yes! let me know if something does not make sense for you... This is just the first step, I think the main go would be to use s2i , configure some pipeline integration with travis and create openshift templates for it. Something like we have here:

* https://github.com/3scale/apicast

EricWittmann:
Great, thanks!

PR #92, opened 2017-03-29
Introducing Local/Remote release deployments

maneta:
In this PR we introduce the possibility to deploy local or remote release to the docker registry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 NAME = apiman-studio
 NAMESPACE = maneta
 RELEASE_VERSION ?= 0.0.5
+LOCAL_RELEASE ?= 'false'
 LOCAL_IMAGE := $(NAME):$(RELEASE_VERSION)
 REMOTE_IMAGE := $(NAMESPACE)/$(LOCAL_IMAGE)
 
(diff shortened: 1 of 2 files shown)


EricWittmann:
Nice.  Just read the docs on ADD vs. COPY for a dockerfile.  Looks like this was easier than you thought it would be eh?  :)

Makefile
```diff
@@ -16,8 +17,15 @@ all: build
 update: build push
 
 build: ## Build docker image with name LOCAL_IMAGE (NAME:RELEASE_VERSION).
-	docker build -f $(THISDIR_PATH)/Dockerfile -t $(LOCAL_IMAGE) $(PROJECT_PATH) \
-		--build-arg RELEASE_VERSION=$(RELEASE_VERSION)
+	@if [ -f "./front-end/quickstart/target/api-design-studio-${RELEASE_VERSION}-quickstart.zip" ] || [ ${LOCAL_RELEASE} == "true" ]; then \
```
maneta: I Could set the RELEASE_PATH variable in the subshell, but doing it replicating the `docker build` command it improves readability

maneta:
yes it was :)

PR #226, opened 2018-01-19
Able to download WildFly and KeCloak adapter with maven

dmvolod:
This reduce a build time and use maven cache for builds.
Also shell permissions inside quickstart also fixed.
Please review.


EricWittmann:
Hey thanks for the PR - this is great!

PR #238, opened 2018-02-22
Set limits on Openshift Deployment Configs

orimarti:
Parametrizable limits on the Openshift Deployment Configs


EricWittmann:
Thanks @orimarti .

PR #316, opened 2018-06-07
Create Angular package exposing ApiEditorComponent

zregvart:
In https://github.com/syndesisio/syndesis we're considering embedding a OpenAPI editor functionality of Apicurio. To do this we would like to depend on a Angular module that contains only the `ApiEditorComponent`.

This is the first step in creating an Angular package that only exposes the `ApiEditorComponent`. More details are provided in the individual commit messages.

The base commit that accomplishes the creation of the package and exposes the `ApiEditorComponent` is in 8402911. This is done by using https://github.com/dherges/ng-packagr/. Other commits resolve errors/warnings emitted from the ahead of time compilation.

Next steps might be to use this `ApicurioEditorModule` internally within Apicurio so that it's always up to date with the dependencies or other code changes to the `ApiEditorComponent`, or to create a test that embeds the `ApiEditorComponent` from the `ApicurioEditorModule` to validate the embedding scenario.

### DIFF
--- a/front-end/studio/package.json
+++ b/front-end/studio/package.json
@@ -12,10 +12,11 @@
     "ng": "ng",
     "start": "ng serve --port 8888 --aot",
     "debug": "ng serve --port 8888 --sourcemap -dev -v",
-    "build": "ng build -prod"
+    "build": "ng build -prod",
+    "package": "ng-packagr -p package.json"
   },
   "private": true,
-  "dependencies": {
+  "peerDependencies": {
     "@angular/animations": "^5.2.0",
     "@angular/common": "^5.2.0",
     "@angular/compiler": "^5.2.0",
@@ -24,22 +25,45 @@
     "@angular/platform-browser": "^5.2.0",
     "@angular/platform-browser-dynamic": "^5.2.0",
     "@angular/router": "^5.2.0",
-    "patternfly": "3.40.0",
-    "oai-ts-core": "0.2.15",
-    "oai-ts-commands": "0.2.28",
+    "brace": "0.11.1",
     "core-js": "^2.4.1",
     "ngx-bootstrap": "2.0.2",
-    "brace": "0.11.1",
-    "yamljs": "0.3.0",
+    "oai-ts-commands": "0.2.28",
+    "oai-ts-core": "0.2.15",
+    "patternfly": "3.40.0",
     "rxjs": "^5.5.6",
+    "yamljs": "0.3.0",
     "zone.js": "^0.8.19"
   },
   "devDependencies": {
+    "@angular/animations": "^5.2.0",
     "@angular/cli": "1.6.8",
+    "@angular/common": "^5.2.0",
+    "@angular/compiler": "^5.2.0",
     "@angular/compiler-cli": "^5.2.0",
+    "@angular/core": "^5.2.0",
+    "@angular/forms": "^5.2.0",
     "@angular/language-service": "^5.2.0",
+    "@angular/platform-browser": "^5.2.0",
+    "@angular/platform-browser-dynamic": "^5.2.0",
+    "@angular/router": "^5.2.0",
     "@types/node": "~6.0.60",
+    "brace": "0.11.1",
+    "core-js": "^2.4.1",
+    "ng-packagr": "^2.0.0",
+    "ngx-bootstrap": "2.0.2",
+    "oai-ts-commands": "0.2.28",
+    "oai-ts-core": "0.2.15",
+    "patternfly": "3.40.0",
+    "rxjs": "^5.5.6",
     "ts-node": "~4.1.0",
-    "typescript": "~2.5.3"
+    "typescript": "~2.5.3",
+    "yamljs": "0.3.0",
+    "zone.js": "^0.8.19"
+  },
+  "ngPackage": {
+    "lib": {
+      "entryFile": "public_api.ts"
+    }
   }
 }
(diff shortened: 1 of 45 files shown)


EricWittmann:
Thanks!  I'll get this merged today.

front-end/studio/package.json
```diff
@@ -12,10 +12,11 @@
     "ng": "ng",
     "start": "ng serve --port 8888 --aot",
     "debug": "ng serve --port 8888 --sourcemap -dev -v",
-    "build": "ng build -prod"
+    "build": "ng build -prod",
+    "package": "ng-packagr -p package.json"
   },
   "private": true,
-  "dependencies": {
+  "peerDependencies": {
```
EricWittmann: Can you explain the changes in package.json and how they will impact apicurio when built as an application (vs. as a module to be used in e.g. syndesis)?

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/dialogs/add-path.component.ts
```diff
@@ -20,7 +20,7 @@ import {ModalDirective} from "ngx-bootstrap";
 import {OasDocument, OasLibraryUtils} from "oai-ts-core";
 import {FormControl} from "@angular/forms";
 import {Subject} from "rxjs/Subject";
-
+import 'rxjs/add/operator/distinctUntilChanged';
```
EricWittmann: Why was this added?  I can't see where it is used...

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/common/inline-editor.base.ts
```diff
@@ -261,7 +261,7 @@ export abstract class TextAreaEditorComponent extends TextInputEditorComponent {
 
     /**
      * Subclasses can override this to provide validation status of the current value.
-     * @return {boolean}
```
EricWittmann: Lots of jsdoc changes in this diff.  I think WebStorm adds this automatically.  I'm happy to remove them all but am wondering why. :)

front-end/studio/src/app/pages/apis/apis.page.ts
```diff
@@ -78,7 +78,7 @@ export class ApisPageComponent extends AbstractPageComponent implements OnDestro
     public allApis: Api[] = [];
     public filteredApis: Api[];
     public selectedApis: Api[];
-    public filters: Filters = new Filters();
+    private filters: Filters = new Filters();
```
EricWittmann: Any particular reason to make this particular field private?  I'm sure there are lots of fields around that could be private but are not.

front-end/studio/package.json
zregvart: So this is my understanding of `dependencies` vs `peerDependencies`: when we use `dependencies` NPM (Yarn) will put the dependencies listed within, whereas with `peerDependencies` it will put them on the same level with Apicurio.

The difference can be seen in `node_modules` when you add Apicurio as a dependency:

Declared as `dependencies`:
```
node_modules
├── apicurio-design-studio
...
│   ├── node_modules
│   │   ├── bootstrap-datepicker
...
│   │   │   ├── node_modules
│   │   │   │   └── jquery
...
│   │   ├── bootstrap-switch
...
│   │   ├── brace
│   │   │   └── index.d.ts
│   │   ├── camel-case
│   │   │   └── camel-case.d.ts
...
```

Declared as `peerDependencies`:
```
├── apicurio-design-studio
...
│   ├── node_modules
│   │   ├── brace
│   │   │   └── index.d.ts
│   │   ├── camel-case
│   │   │   └── camel-case.d.ts
...
```

Notice that we then don't get `bootstrap-datepicker` or `bootstrap-switch` (and so on) dependencies in the project that uses Apicurio.

If you try to declare dependencies as `dependencies`, you get an error from ng-packagr:
```
Distributing npm packages with 'dependencies' is not recommended. Please consider adding @angular/animations to 'peerDependencies' or remove it from 'dependencies'.

BUILD ERROR
Dependency @angular/animations must be explicitly whitelisted.
Error: Dependency @angular/animations must be explicitly whitelisted.
```

This is somewhat similar to optional dependencies when using Maven.

front-end/studio/src/app/pages/apis/apis.page.ts
zregvart: I can't remember exactly, but I think the ahead of time compilation failed as I haven't imported `Filters` in the `public_api.ts`, being that `Filters` is only used here made more sense to make it `private`.

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/common/inline-editor.base.ts
EricWittmann: Yeah figured that one out.  WebStorm adds the {Type} automatically ATM, but the next patched version of it will add an option to toggle that behavior (and set the default for typescript to false).  So that's good. :)

I'll go through all of Apicurio at some point and remove them.

zregvart: You could do `find . -name '*.ts' -exec sed -i -e 's/\(@param\|@return\) {[^}]*}/\1/' {} \;` or some other bulk magic.

EricWittmann: +1 - I do that sort of thing with regular expressions from within Eclipse.  It has nice support for find/replace across files using regexp and with full preview before approving the change.

EricWittmann: FWIW: https://youtrack.jetbrains.com/issue/WEB-28964

PR #434, opened 2018-09-20
Stop context menu disappearing on macOS

### LINKED ISSUE #433: Editor: Context menu not working properly in MacOS + Firefox
The `@HostListener("document:click", ["$event"])` handler is being invoked immediately after the context menu is shown.

msavy:
Attempts to fix #433

### DIFF
--- a/front-end/studio/src/app/pages/apis/{apiId}/editor/_components/master.component.ts
+++ b/front-end/studio/src/app/pages/apis/{apiId}/editor/_components/master.component.ts
@@ -493,7 +493,12 @@ export class EditorMasterComponent implements OnInit, OnDestroy {
      * menu if it is open.
      */
     @HostListener("document:click", ["$event"])
-    public onDocumentClick(): void {
+    public onDocumentClick(event: MouseEvent): void {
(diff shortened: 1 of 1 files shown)


msavy:
Please review, @EricWittmann

msavy:
There's a chance this could be too blunt an instrument, in which case will have to think of another solution. But, I can't /see/ any side effects so far.

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/master.component.ts
```diff
@@ -493,7 +493,12 @@ export class EditorMasterComponent implements OnInit, OnDestroy {
      * menu if it is open.
      */
     @HostListener("document:click", ["$event"])
-    public onDocumentClick(): void {
+    public onDocumentClick(event): void {
```
EricWittmann: Can you add the type to the `event` parameter?  Should be `event: MouseEvent`.

msavy:
Your suggestion should be applied now, hopefully.

PR #446, opened 2018-09-24
Fixes Syndesis Issue 3497: Apicurio GUI - info box not showed properly

elvisisking:
- when info box extends below bottom of window user can now scroll the window to see entire contents
- removed `left` and `right` properties from the `ContextHelpComponent`
- info box positioning is all  done in the css now


EricWittmann:
Thanks!

PR #476, opened 2018-10-15
Add macOS shortcuts for undo/redo

msavy:
Want to do a bit more testing on this first before merging, but had a couple of questions (inline)

### DIFF
--- a/front-end/studio/src/app/pages/apis/{apiId}/editor/editor.component.ts
+++ b/front-end/studio/src/app/pages/apis/{apiId}/editor/editor.component.ts
@@ -193,12 +193,22 @@ export class ApiEditorComponent implements OnChanges, OnInit, OnDestroy, IEditor
      */
     public onGlobalKeyDown(event: KeyboardEvent): void {
         // TODO skip any event that was sent to an input field (e.g. input, textarea, etc)
-        if (event.ctrlKey && event.key === 'z' && !event.metaKey && !event.altKey) {
+        if (ApiEditorComponent.isUndo(event)) {
             this.undoLastCommand();
         }
-        if (event.ctrlKey && event.key === 'y' && !event.metaKey && !event.altKey) {
+        if (ApiEditorComponent.isRedo(event)) {
             this.redoLastCommand();
         }
+    }
+
+    private static isUndo(event: KeyboardEvent): boolean {
+        return (event.ctrlKey && event.key === 'z' && !event.metaKey && !event.altKey) ||
+            (event.metaKey && event.key === 'z' && !event.ctrlKey && !event.altKey);  // macOS: cmd + z
+    }
+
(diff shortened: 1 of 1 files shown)


front-end/studio/src/app/pages/apis/{apiId}/editor/editor.component.ts
```diff
@@ -193,14 +193,24 @@ export class ApiEditorComponent implements OnChanges, OnInit, OnDestroy, IEditor
...
+        if (ApiEditorComponent.isUndo(event)) {
             this.undoLastCommand();
         }
-        if (event.ctrlKey && event.key === 'y' && !event.metaKey && !event.altKey) {
+        if (ApiEditorComponent.isRedo(event)) {
             this.redoLastCommand();
         }
     }
 
+    private static isUndo(event: KeyboardEvent): boolean {
```
msavy: Would you prefer these in a utility/helper class somewhere? If so, where?

front-end/studio/src/app/pages/apis/{apiId}/editor/editor.component.ts
```diff
@@ -193,14 +193,24 @@ export class ApiEditorComponent implements OnChanges, OnInit, OnDestroy, IEditor
...
+        return (event.ctrlKey && event.key === 'z' && !event.metaKey && !event.altKey) ||
+            (event.metaKey && event.key === 'z' && !event.ctrlKey && !event.altKey);  // macOS: cmd + z
```
msavy: My thoughts on this was that we support both options (macOS shortcuts and windows/linux). So you can do it the Linux/Windows way on Mac if you want.

front-end/studio/src/app/pages/apis/{apiId}/editor/editor.component.ts
msavy: Also, leaving them here seems completely reasonable to me, but just wanted to check :)

EricWittmann:
Looks good to me - but I've read that not all browsers necessarily support the "meta" flag on the event.  Not sure if that's still the case - so if you could just test whatever major browsers you have, that'd be great (presumably FF, Chrome, Safari... and obviously IE for Mac!).

PS: kidding about IE for mac

msavy:
Generally seems good— https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey#Browser_compatibility

Will try it on mobile if my half-dead phone will tolerate it LOL.

msavy:
I've tested this and it seems to work well.

I forgot to sign the commit, will can push a signed version if you wish.

EricWittmann:
It's not required, just recommended as best practice.  No worries.

msavy:
Yes, I normally have it, just forgot to set it up for this repo. 

Although, if you merge via the web UI it generally destroys the signature anyway!

Issue #485, opened 2018-11-01
OpenShift:  Split the template into two (separate the auth component from everything else)


odra:
Hello,

I am currently developing an apicurio openshift operator using [this repo's template](https://github.com/Apicurio/apicurio-studio/blob/master/distro/openshift/apicurio-template.yml) it will use an existing keycloak instance rather than deploying a new one.

- I would to prefer to use the upstream template as it is without any modifications, would it be possible to provide a template in the future which does not deploy an auth server so it will just use an existing keycloak instance by providing its auth url?
-  What happens if I try to use an existing keycloak instance which already contains the realm to be used by apicurio? Would it use the existing realm?

 cc @david-martin

EricWittmann:
Sure - we can just split the existing template into two.  That's no problem.

And yes - it's easy enough to configure Apicurio to use an existing Keycloak.  However, Apicurio does require some specific configuration within the Keycloak realm in order for all Apicurio features to work.

EricWittmann:
If interested, more information about configuring Keycloak to work with Apicurio can be found here:

https://apicurio-studio.readme.io/docs/setting-up-keycloak-for-use-with-apicurio

odra:
Ah, thanks for the doc link :)

Another question, I downloaded the source code from the latest release link (0.2.18-final) and most images in the template are using the "latest-release" tag, should the template be using the "0.2.18.Final" one for all related apicurio container images instead?

EricWittmann:
Hm.  The openshift template was originally created for https://studio.apicur.io/ - which auto-updates to the latest version whenever I do a release.  In that context, `latest-release` is obviously the right tag.  In another context it might not be.  I guess I'm not sure - it probably depends on the use-case.

odra:
Basically what I was looking for is to lock a specific version in the template, it could happen that a deployment is using the template file from 0.2.18-final but ends up pulling the image of an earlier release (let's say 0.2.19) since it is always using the latest-release tag (I know I can change it myself, but I am trying to not change anything in the template).

Around splitting the template into two, do you manually do it or is there some kind of script that generates the template? Happy to send a pr around that :)

EricWittmann:
The template is maintained manually - so a PR would be most welcome. :)

[...]

PR #524, opened 2018-12-11
openshift templates

odra:
Adding split templates:

- postgresql
- auth
- studio (ui + ws + api)

I was originally just splitting  the auth one (keycloak) but it turns out that we neeed to deploy postgre first so I created a postgre template as well.

The "full featured" template still exists, this change is being made based on this issue comments: #485

### DIFF
--- a/distro/openshift/apicurio-auth-template.yml
+++ b/distro/openshift/apicurio-auth-template.yml
@@ -1,0 +1,259 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: apicurio-auth
+message: |-
+  Congratulations on deploying Apicurio Studio into OpenShift!
+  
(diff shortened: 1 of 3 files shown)


odra:
Thanks! :+1:

I can help with any "release scripting" if needed :D

EricWittmann:
I have updated the release script **and** I have done a new release!  So the latest Apicurio release is now working as we have designed.  :)  If you look at the Apicurio OpenShift template for the latest release here you will see the updates:

https://github.com/Apicurio/apicurio-studio/blob/v0.2.19.Final/distro/openshift/apicurio-template.yml

Whereas the same file on the `master` branch uses `latest-release` as the tag rather than `0.2.19.Final`.  @odra - let me know if there's anything I've missed.

Issue #639, opened 2019-02-06
Update README with quickstart instructions.


msolberg:
It took me a minute to figure out how to start the quickstart from the source tree. Here's an updated README.

EricWittmann:
Fixed by https://github.com/Apicurio/apicurio-studio/pull/640

PR #640, opened 2019-02-06
Updating readme for #639

msolberg:
Hi. Here's a quick patch for #639.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,13 @@ The following command compiles all the code, installs the JARs into your local M
 ```
 ## Quickstart (i.e. How To Run It)
 
-TBD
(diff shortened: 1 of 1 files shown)


msolberg:
Actually, now I'm having trouble getting the application up and running - let me spend some more time on this.

EricWittmann:
Yeah, that directory you've referenced is just a work dir for assembling the final .zip file.  What really needs to happen is to copy the zip file found in that `target` directory, unpack it, and then start it up with the command you listed.

EricWittmann:
I've updated the documentation and will merge once the circleci check passes. :)

PR #676, opened 2019-03-05
Empty inline text editor should hide the "clear" button. Fixes #607

### LINKED ISSUE #607: Editor: Inline text editor should hide the (x) when content is already empty
There is a shared inline text input control which has a little (x) that you can click on to clear out the value of the field.  It looks like this:

image

When there is no value in the field, that (x) button should not be visible.  Here is what it looks like now:

image

But this should be changed so that the (x) button is missing when the value is empty.

jsenko:


front-end/studio/src/app/pages/apis/{apiId}/editor/_components/common/inline-editor.base.ts
```diff
@@ -168,7 +175,6 @@ export abstract class TextInputEditorComponent extends AbstractInlineValueEditor
         this.input.changes.subscribe(changes => {
             if (changes.last) {
                 changes.last.nativeElement.focus();
-                changes.last.nativeElement.select();
```
jsenko: Is this important? When starting with an empty text field, it caused the first character typed to be immediately selected, and overwritten with the next one.

EricWittmann: I think auto selecting the value when there is a value is a nice feature.  Instead of removing that line can you check for a value and skip the line of the value is empty?

jsenko: Done!

EricWittmann: +1

PR #677, opened 2019-03-05
[in-progress] Refactor inline editors

jsenko:
@EricWittmann Hi, Could you please add an `Work In Progress` / `Experimental` / `Do not integrate yet` label for PRs?

I was playing with the code after #676 and noticed possible improvements. If this looks like a good way to go, I can finish it.

```
 - Reduce code duplication
 - Refactor class hierarchy
```


EricWittmann:
This is a great area for improvement and this looks like a great start.  There was originally a sensible class hierarchy for inline editors but UX changes prompted a small rewrite and I lost a fair bit of the reuse...and never went back in to improve it.  So this is great!

jsenko:
I'm going to close this as outdated. It could be revisited later.

PR #679, opened 2019-03-06
Support "file" as an option for OpenAPI 2 form body parameters

jsenko:
@EricWittmann Hi, this is a WIP, I couldn't figure out how to propagate the "file" type into the OpenAPI model. The use and processing of the `SimplifiedType` is not very clear to me atm.


EricWittmann:
Ok no problem we can go over it when I get back next week.

PR #703, opened 2019-03-26
Implement "Clone Response" action

jsenko:
Merge after https://github.com/Apicurio/oai-ts-commands/pull/46

@EricWittmann Hey, there is an issue I couldn't yet solve - if you clone a response in UI, 
everything works OK, but if you reload the page, the content of the cloned response
is missing, e.g. the description is null. Any advice?

There is also another minor problem, when you create/clone a response and switch to its tab,
after hitting "undo", the tab disappears, but the UI does not switch to another one.
The simple solution for the delete action does not seem to directly applicable.

There are also some minor code improvements, that can be separated out if need be.

### DIFF
--- a/front-end/studio/src/app/pages/apis/create/_components/create-form.component.ts
+++ b/front-end/studio/src/app/pages/apis/create/_components/create-form.component.ts
@@ -17,7 +17,7 @@
 
 import {Component, EventEmitter, Inject, Output} from "@angular/core";
 import {IAuthenticationService} from "../../../../services/auth.service";
-import {DropDownOption} from '../../../../components/common/drop-down.component';
+import {DropDownOption, DropDownOptionValue as Value} from '../../../../components/common/drop-down.component';
 import {ApisService} from "../../../../services/apis.service";
 import {LinkedAccountsService} from "../../../../services/accounts.service";
 import {TemplateService} from "../../../../services/template.service";
@@ -68,8 +68,8 @@ export class CreateApiFormComponent {
 
     public typeOptions(): DropDownOption[] {
         return [
-            { name: "Open API 2.0 (Swagger)", value: "2.0"},
-            { name: "Open API 3.0.2", value: "3.0.2"}
+            new Value("Open API 2.0 (Swagger)", "2.0"),
+            new Value("Open API 3.0.2", "3.0.2")
         ];
     }
 
--- a/front-end/studio/src/app/pages/apis/{apiId}/editor/_components/dialogs/add-response.component.ts
+++ b/front-end/studio/src/app/pages/apis/{apiId}/editor/_components/dialogs/add-response.component.ts
@@ -19,133 +19,8 @@ import {Component, EventEmitter, Output, QueryList, ViewChildren} from "@angular
 import {ModalDirective} from "ngx-bootstrap";
 import {Oas20Operation, Oas30Operation} from "oai-ts-core";
 import {DropDownOption, DropDownOptionValue as Value, DIVIDER} from "../../../../../../components/common/drop-down.component";
+import {HttpCodeService} from "../../_services/httpcode.service";
 
-export class HttpCode {
-
-    private code: number;
-    private name: string;
-
-    public constructor(code: number, name: string) {
-        this.code = code;
-        this.name = name;
-    }
-
-    public getCode(): number {
-        return this.code;
-    }
-
-    public getName(): string {
-        return this.name;
-    }
-}
-
-export class HttpCodeUtil {
-
-    private static HTTP_CODE_LIST_COMMON: HttpCode[] = [
-        new HttpCode(200, "OK"),
-        new HttpCode(201, "Created"),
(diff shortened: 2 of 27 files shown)


EricWittmann:
Don't worry about the problem of reloading the browser.  The reason for that is because the server is sometimes responsible for executing the commands - and until a **release** of `oai-ts-commands` is performed, the server won't know about any new (or updated) commands.  So this problem will get resolved once that release is done (the release process for `oai-ts-commands` automatically updates the apicurio studio server).

EricWittmann:
As for the tab+undo problem - we should open up a separate GH issue for that.  We'll have to think about that - I'm guessing that this is a more general problem already (create something, select it, then **undo** it).

front-end/studio/src/app/pages/apis/create/_components/create-form.component.ts
```diff
@@ -68,8 +68,8 @@ export class CreateApiFormComponent {
 
     public typeOptions(): DropDownOption[] {
         return [
-            { name: "Open API 2.0 (Swagger)", value: "2.0"},
-            { name: "Open API 3.0.2", value: "3.0.2"}
+            new Value("Open API 2.0 (Swagger)", "2.0"),
+            new Value("Open API 3.0.2", "3.0.2")
```
EricWittmann: I was unsure about this change (adding methods to the class) but this is pretty nice. :)

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/dialogs/add-response.component.ts
```diff
@@ -18,7 +18,134 @@
...
+    public getCode(): number {
+        return this.code;
+    }
+
+    public getName(): string {
+        return this.name;
+    }
+}
+
+export class HttpCodeUtil {
```
EricWittmann: Can you leverage this here?

https://github.com/Apicurio/apicurio-studio/blob/master/front-end/studio/src/app/pages/apis/%7BapiId%7D/editor/_services/httpcode.service.ts

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/dialogs/add-response.component.ts
```diff
@@ -18,7 +18,134 @@
...
+        new HttpCode(409, "Conflict"),
+        new HttpCode(410, "Gone"),
+        new HttpCode(411, "Length Required"),
+        new HttpCode(412, "Precondition Failed"),
+        new HttpCode(413, "Payload Too Large"),
+        new HttpCode(414, "URI Too Long"),
+        new HttpCode(415, "Unsupported Media Type"),
+        new HttpCode(416, "Range Not Satisfiable"),
+        new HttpCode(417, "Expectation Failed"),
+        new HttpCode(418, "I'm a teapot!"),
```
EricWittmann: This code should only be in the list if the current date is April 1st. :)  

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/dialogs/clone-response.component.html
```diff
@@ -0,0 +1,36 @@
+<!-- Add Response Dialog -->
```
EricWittmann: Clone Response Dialog

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/dialogs/clone-response.component.ts
```diff
@@ -0,0 +1,136 @@
+/**
+ * @license
+ * Copyright 2017 JBoss Inc
```
EricWittmann: 2019

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/dialogs/add-response.component.ts
EricWittmann: Nevermind - I see that you addressed this in the TS file.  Nice. :)

front-end/studio/src/app/pages/apis/{apiId}/editor/_components/forms/path/operation/responses-section.component.html
```diff
@@ -36,6 +36,7 @@
                     </button>
                     <ul class="dropdown-menu dropdown-menu-right">
                         <li><a (click)="deleteResponse(currentResponse())"><span class="fa fa-fw fa-trash"></span><span>Delete Response</span></a></li>
+                        <li><a (click)="openCloneResponseModal(currentResponse())"><!--<span class="fa fa-fw fa-trash"></span>--><span>Clone Response</span></a></li>
```
EricWittmann: Use the clone/copy icon - same as **Clone Path**.

Issue #700, opened 2019-03-20
Create standalone docker distribution (based on docker compose)


EricWittmann:
We have the docker images to support a non-openshift based (docker compose) distribution.  Such a distro would be easier to both evaluate and deploy for users who do not have an OpenShift cluster available.

EricWittmann:
My initial thought is to add a new folder in the `distro` section of the codebase:

https://github.com/Apicurio/apicurio-studio/tree/master/distro

It can contain the docker compose file, a Readme, and anything else needed to run that I'm not aware of. :)

chriske:
Hey!

On the other issue (the WildFly), you asked me if I would like to contribute with the All-In-One compose bundle with Microcks. For some reason I didn't get the notification, but I'm here right now. When Nicnl contributes with the compose file, I can extend it to that All-in-one bundle with Microcks and KeyCloak.  Let me now if you want this :) And after it's done we can contribute that modification to Microcks too.

Nicnl:
Hey  
Sorry for the delay, I don't have access to those compose files from home

I created a gist with two compose files:  
https://gist.github.com/Nicnl/652e70e7ba364cbb196fe4cdc5bd622d

- One for Keycloak and its database  
(Sorry Eric, I used the official `jboss/keycloak` container instead of your `apicurio-studio-auth`)
- One for Apicurio Studio

I stripped them as much as I could removing most of my environment-specific settings, hopefully keeping the absolute minimum  

However, there's a few points that are making me quite unconfortable:
- **The MySQL database**  
I used a MySQL database both for Keycloak and Apicurio  
But by doing so I encountered multiple issues, notably #683 and #688  
Could it be wiser to modify it and use a Postgres database instead?  
It's also possible to provide multiple _flavors_ of the compose files, giving the user the choice of picking his favorite database.  
I don't know.

- Even if I tried to strip those compose to their bare minimum, there's a few environment specific settings that I'm not able to remove:  
  - **The port mappings**  
    For now, I used the ports 8090 for Keycloak, 8091 for the API, 8092 for the websockets and 8093 for the UI  
    Those are hardcoded

  - **The environment variables of the UI container**  
    The UI container required the full endpoint URLs  
    My local server being `10.0.0.130`, I've configured this:  
      - `APICURIO_KC_AUTH_URL: 'http://10.0.0.130:8090/auth/'`
      - `APICURIO_UI_HUB_API_URL: 'http://10.0.0.130:8091'`
[...]

PR #707, opened 2019-03-30
#700 added docker-compose distro without reverse proxy and SSL support

chriske:


chriske:
#700

EricWittmann:
Taking a look at this now, sorry for the delay.

EricWittmann:
This is a fantastic start - I tried it and it worked great.  There are some improvements I think we can/should make and I'm not 100% sure how to address them all.  So here is some feedback and we can discuss (in no particular order):

1) I can't log in to Microcks and there is no **Register User** option
2) The Apicurio version is hard-coded and should either be parameterized in some way or else set to "latest-release" and then modified to a specific version whenever a release is done (similarly to what is done now for the OpenShift templates)
3) Missing the Keycloak login theme for Apicurio (i.e. getting the classic Keycloak theme rather than the Apicurio theme for the login page)
4) Logout is broken (the logout servlet is redirecting to `/studio` and should be configured)

EricWittmann:
The issue with (1) is just realm configuration - so we can update the microcks realm file to enable registration.

EricWittmann:
For (4) we need to introduce the following environment variable:

`APICURIO_UI_LOGOUT_REDIRECT_URI`

And its value should be:

`/`

EricWittmann:
I guess for (3) we would need a custom image to install the theme.  For reference the Apicurio KC login theme is here:

https://github.com/Apicurio/apicurio-studio/tree/master/distro/keycloak/themes-kc3/apicurio

EricWittmann:
So @chriske - thoughts on the other items in the list?  I'm actually happy accepting this PR as-is and we can address some of these issues iteratively if you'd prefer.

chriske:
@EricWittmann Thank you for your feedback! I totally agree with your improvements. I think I'm going to extend this PR with (1)-(4). I also want to extend the Readme with some configuration part and explanation.

(5) and (6) are a little bit more complex, but we can do them later, it's not an issue.  To make (5) possible, I will introduce a new compose and .env file. For (6) I think I have a really nice solution, and we can extend the setup procedure with password generation too (and we should also generate the api key for microcks too).  This can be the 2nd iteration of the task, and the last one will be the reverse proxy and SSL support.

chriske:
I finished the modifications:

(1) - modified Microcks realm config, users can now register directly
(2) - made the versions configurable in the .env file, the default docker tag is: `latest-release`. Also added a new section about this into the Readme
(3) - added a build step, which will create an own keycloak image based on the context of the compose file. This means I had to extend the starting steps in the Readme. I created a keycloak folder in the docker-compose folder, and that folder contains the new Dockerfile and your theme. Unfortunately I can't link it to the already existing folder in dist, because it has to be inside the docker-compose folder :(
(4) - added the new ENV variable with your default value

If you tried it, and you think it's good enough, merge it please. We will do the next steps with new issues/PRs if that's OK for you.

Thanks!

EricWittmann:
Thanks for doing those updates, @chriske - I'm hopeful I will be able to re-test this today or tomorrow and then get it merged.

EricWittmann:
Hey @chriske this is all really great.  All my testing was successful.  I look forward to future improvements!  But what an amazing start, and it's such a great way for docker users to get going quickly.  I'm going to update the project website to include instructions for running this way (probably just link to your readme).

chriske:
Hey @EricWittmann thanks for your kind words! You're welcome!

PR #722, opened 2019-04-10
Initial integration test for single node apicurio with keycloak. Fixes #702

### LINKED ISSUE #702: Create integration tests with Keycloak
Create integration tests with Keycloak to make sure we can update to the newer versions without issues [1]. Use JMS integration test [2] as a starting point.

[1] https://github.com/Apicurio/apicurio-studio/pull/680
[2] https://github.com/Apicurio/apicurio-studio/blob/master/test/integration/arquillian/src/test/java/io/apicurio/test/integration/arquillian/VerifyDistributedSetupTestIT.java

jsenko:


EricWittmann:
Should we be leveraging the `ProcessExecutor` class in the other integration test?

Issue #630, opened 2019-01-30
Editor: Send updated selection event when adding/creating an entity


EricWittmann:
When the user either creates an entity or perhaps **starts creating** an entity, we should fire a selection event so that collaborators can see what the user is doing.

jsenko:
@EricWittmann Could you please provide more details to the description?

PR #740, opened 2019-04-18
Emit selection events when creating new document nodes / entities. Fi…

jsenko:
…xes #630

This should work, I think I caught all places where new OA nodes / entities are created,
but I'm not entirely happy with the straightforward implementation.
I'm thinking of placing the responsibility of getting the paths into the commands.
Also, working with the paths is not ideal IMHO, I can maybe leverage the type system to make creating them easier.

### DIFF
--- a/front-end/studio/src/app/pages/apis/{apiId}/editor/_components/common/base-component.ts
+++ b/front-end/studio/src/app/pages/apis/{apiId}/editor/_components/common/base-component.ts
@@ -18,13 +18,10 @@ import {ChangeDetectorRef, OnChanges, OnDestroy, OnInit, SimpleChanges} from "@a
 import {DocumentService} from "../../_services/document.service";
 import {SelectionService} from "../../_services/selection.service";
 import {TopicSubscription} from "apicurio-ts-core";
-import {OasLibraryUtils} from "oai-ts-core";
 
 export abstract class AbstractBaseComponent implements OnInit, OnChanges, OnDestroy {
 
     private _docSub: TopicSubscription<void>;
-
-    protected static oasLibrary: OasLibraryUtils = new OasLibraryUtils();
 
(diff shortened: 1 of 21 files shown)


EricWittmann:
So I think this would be improved if you moved some of the logic into the selection service itself.  Perhaps have a variant (or perhaps a small number of variants) of `select` that are more convenient for this use-case.  So for example,  you could add the following:

```typescript
    public selectNode(node: OasNode, ...additionalSegments: string[]): void;
```

Or something similar.

The difficulty in working with paths in this case is primarily because the entity being created hasn't YET been created.  So you can't simply create a path from a node.  Instead you have to create the path and then augment it (as you have done).  But again, a simple convenience method on the selection service takes care of the headaches I think (with the possible exception of the indexed paths like for parameters).

Note that I've long been considering an enhancement to NodePath that would support an optional syntax for selecting items within an array by property rather than by index.  So for example right now we only support this syntax for selecting a Tag Definition:

`/tags[2]`

This would select the 3rd tag definition in the document.  However, a better syntax would be:

`/tags[name=Some Tag Name]`

Some nice things about that:

* Paths can more easily be constructed before the entity itself has been created
* Paths are more canonical - they would work even if the user re-ordered the tag definitions
* Better for unit testing due to less fragility in the paths
* Better for auditing - the path makes more sense

Some pitfalls to supporting this:

* Parsing the path will be harder because we'll need to worry about weird entity names
* The path breaks if the tag definition is renamed (this is probably not a big deal)
* Might be impossible to use in some situations (e.g. if an entity is a `$ref ` rather than an inline entity)

jsenko:
@EricWittmann Hi, I've rebased and updated the PR. I though about adding the method to the selection service as we discussed, but ultimately I don't think it'll have a big impact and would suggest merging the PR as it is for now.

PR #869, opened 2019-08-01
Adding parameters examples management

lbroudoux:
This commit adds examples management for `path`, `query` and `header` parameters.

It requires updating dependencies to `apicurio-data-models:1.0.8-SNAPSHOT` once https://github.com/Apicurio/apicurio-data-models/pull/49 will be accepted and merge (this later PR adds necessary commands for adding/updating/removing example to a parameter).


EricWittmann:
This is great, thanks @lbroudoux !!

Issue #833, opened 2019-07-02
Upgrade to Keycloak 7


EricWittmann:
We are now two major versions behind.  We should upgrade to Keycloak 6.  What needs to be done to accomplish this:

* Upgrade the keycloak version in the parent `pom.xml`
* Upgrade the `apicurio-studio-auth` docker image to use KC 6
* Make sure the Apicurio login theme works with KC 6 and update accordingly
* Make sure that https://github.com/Apicurio/apicurio-keycloak-extensions works with KC 6
* TEST TEST TEST - especially identity providers (github, gitlab, bbucket)

asheliahut:
I have been running keycloak 6 with docker and able to run fine with the apicurio theme. Just wanted to add that to maybe bump the work being done to just knock this out.

asheliahut:
@EricWittmann Just to note keycloak 7 is out now.

EricWittmann:
My clever plan of being too busy to deal with Keycloak 6 has finally paid off!  Waa ha ha ha!

OK seriously though - I did see this and I do need to get Apicurio upgraded to KC 7.  I very much hope to find bandwidth for this soon.

Criptos:
New to apicuro, new to keycloak, but really stubborn, any quick go to or doc you would recommend to get up to speed? @asheliahut  ? @EricWittmann  ?  I used the real.json from ./distro/openshift/auth/realm.json but no luck...

EricWittmann:
There is some documentation here:  https://apicurio-studio.readme.io/docs/setting-up-keycloak-for-use-with-apicurio

That might be a decent thing to get started.  It won't require openshift or k8s or anything like that.  If that's not what you're looking for, let me know. :)

Criptos:
@EricWittmann  I come from that documentation... I actually setup KeyCloak 7, 4 is too old and documentaion doesn't seems to apply. 
 
I got the apicuro downloaded and the stand alone doesn't work because the dependancy of keycloak is missing, so I went out and added keycloak just to find that documentation about keycloak is to old, so I went  read here and there but then... you must know the architectural aspects of apicuro, and haven't found how exactly the relation between apicuro and keycloak is... to know what to configure at keycloak...  I would love a more up to date keycloak-apicuro documentation...

EricWittmann:
[...]

PR #950, opened 2019-10-25
Upgrade to Keycloak 7

EricWittmann:
Hey @asheliahut - if you'd like to take a quick look at the changes in this PR and provide any feedback you may have, that'd be great. :)  I think everything is working fine, but another set of eyes would be great.

### DIFF
--- a/distro/openshift/auth/Dockerfile
+++ b/distro/openshift/auth/Dockerfile
@@ -1,4 +1,4 @@
-FROM jboss/keycloak:4.1.0.Final
+FROM jboss/keycloak:7.1.0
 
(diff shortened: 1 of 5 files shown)


asheliahut:
@EricWittmann tbh I am not super familiar with the configuration values here that you did for the upgrade. I have tested that keycloak 7 is indeed working with apicurios latest version though from the docker upgrade alone, but that is as far as my known knowledge here can help. Also the code section outside of the xml configs looks solid too that I understand.

EricWittmann:
Thanks for taking a look, @asheliahut

EricWittmann:
Fixes https://github.com/Apicurio/apicurio-studio/issues/833

PR #962, opened 2019-11-04
Adding basic support for GraphQL files

EricWittmann:


front-end/studio/src/app/pages/apis/{apiId}/api-detail.page.html
```diff
@@ -69,25 +78,31 @@ <h1 [title]="apiIconTitle()" class="card-pf-title">
...
+                                                <li *ngIf="actionEnabled('download-json')">
                                                     <a href="{{ uiUrl }}download?type=api&format=json&id={{ api.id }}" download="{{ api.name }}.json">
                                                         <span class="fa fa-fw fa-download"></span>
                                                         <span>Download (JSON)</span>
                                                     </a>
                                                 </li>
+                                                <li *ngIf="actionEnabled('download-sdl')">
+                                                    <a href="{{ uiUrl }}download?type=api&format=sdl&id={{ api.id }}" download="{{ api.name }}.graphql">
+                                                        <span class="fa fa-fw fa-download"></span>
+                                                        <span>Download (SDL)</span>
```
wtrocki: Side note - GraphQL will have 2 forms:

- Human readable SDL language
- Introspection.json file 

SQL is used in server-side
Introspection is used in client-side generators.
While SDL is perfect here there can be separate issue to download introspection that community can do.

EricWittmann: Yeah I have a TODO in Apicurio to support the JSON format, but I didn't immediately see how to get a JSON formatted document out of the `graphql-java` library I'm using.

wtrocki:
Hi

Sorry for late review. I have installed all tools on my machine and verified the changes. Forgot to write update as I have played with the editor.

Most of the graphql editors focus on docouments which means that they are not suitable for schemas.

Generally there are two paths we can take
- Go with something like ace editor or vscode embedded into source code - This will need to be decision for entire project as. This will allow supporting json yaml or graphql

example https://github.com/graphql-editor/graphql-editor/blob/master/src/editor/code/ace/graphqlschema.ts

Alternatively we can try to hack graphiql with language server.

EricWittmann:
OK interesting.  We're already using the Ace editor - so perhaps a quick win could be to enable SDL support in our existing Ace editor support.

wtrocki:
Ace editor has graphql schema support.
https://github.com/ajaxorg/ace/commit/4cfeb90e96b0fedb5e5b345b33e6f92830685dd5

wtrocki:
Actually this will be quite simple fix.
I haven't checked https://github.com/Apicurio/apicurio-studio/blob/4068016bad5bf66c088628d6d08f7206f9b7768b/front-end/studio/src/app/components/common/code-editor.component.ts 

We can create graphql schema editor component and simply use graphql mode for ace.

PR #985, opened 2019-11-24
Apicurio 891 - Kubernetes configurations

t-rap:
This Pull Request adds basic kubernetes configs.

It is separated into 
4  Deployment configs (-api -ws -ui -db)
1 configmap
1 secret config
1 example persistentvolumeclaim
1 Service Config
1 Ingress Config
1 Readme

As of now it should not be used in production, as there is no standard way of exposing the proper paths. The user can choose between  editing the services to use NodePorts or edit the Ingresses to suit its needs. Either way it needs to be reviewed to be set productive.

### DIFF
--- a/distro/kubernetes/README.md
+++ b/distro/kubernetes/README.md
@@ -1,0 +1,93 @@
+# Kubernetes Installation
+
+## Prerequisites
+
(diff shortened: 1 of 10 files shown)


EricWittmann:
Thanks @t-rap - I'll have a look at this asap!

PR #987, opened 2019-11-26
docker-compose distro keycloak 7.0.1

Yuutakasan:
I also updated the master branch


EricWittmann:
Right now Keycloak 7.0.1 doesn't work with MySQL, so we are going to stick with 7.0.0 for now.  Unless you've been able to get it working?

Yuutakasan:
Excuse me. I sent a PR because I thought it was a commit failure. I will test it.

PR #1009, opened 2019-12-30
docs: Removed + from jvm requirement

wtrocki:


wtrocki:
My bad for failing to understand what plus means

PR #1078, opened 2020-02-26
Fixes a capitalization bug when generating a REST resource.

vinceve:


EricWittmann:
Thanks for the contribution.  This seems like an improvement for sure.  Can you provide some examples her of Data Type names that work better now?

Also, I'm wondering if perhaps we should *actually* capitalize the data type name for the `operationId`.  I think what you've done here is actually pluralized whatever the original name of the data type might be.  So if I type in "mouse" for the data type name, I think I'll get **getmouse** as the `operationId` value, no?  I'm wondering if **getMouse** would pretty much always be better...

vinceve:
Hey Eric,

I was using the application and noticed when I defined a resource ‘Tank’ (for example) every rest operationid has capitalized names except for the collection one. :-) So I always needed to fix that operationId. 

For your second point, I would think you are right. Now it depends on what the user inputs, but it would be nicer to have capitalization out of the box!

EricWittmann:
Got me on the plural of "mouse"!  :) 

Thanks for the feedback.  I'll merge and then do some extra testing to see if I can make it even a little bit more bullet-proof.

PR #1115, opened 2020-04-15
Inital Kafka editing session support.

alesj:


alesj:
Will create new more generic events based support ... so closing this one.

PR #1133, opened 2020-04-29
Events based editing session

alesj:
Adding Kafka and Infinispan support.


EricWittmann:
Looks good!  And initial testing seems good.  I'm going to merge this now and find some time in the future for more thorough testing, with the eventual goal of making the Kafka implementation the official one used in the cloud application.

PR #1135, opened 2020-04-29
Improve quarkus platform documentation

carlesarnal:


platforms/quarkus/api/README.md
```diff
@@ -13,18 +13,56 @@ mvn quarkus:dev
...
+### Build Options
+
+ - `-Pprod` enables Quarkus's *prod* configuration profile, which uses configuration options suitable for a production environment, 
+   e.g. a higher logging level.
+   
+
+##Runtime Configuration
+
+- In the *dev* mode, the application uses an in-memory H2 database`.
+- In the *prod* mode, you have to provide connection configuration for a PostgreSQL server as follows:
```
EricWittmann, line 33: PostgreSQL or MySQL, right?  Both are supported?

EricWittmann: If yes, then we should document how to configure the database type.

platforms/quarkus/api/README.md
```diff
@@ -13,18 +13,56 @@ mvn quarkus:dev
...
+To see additional options, visit:
+ - [OIDC options](https://quarkus.io/guides/security-openid-connect) 
+ 
+ 
+
+
+Note that in order to have the full `apicurio-studio` app running both the `api` and the `ws` needs to share the same database.
+
 
 ## Creating a native executable
```
EricWittmann, line 60: Let's drop this section - I don't think we're ever going to be building a native image.

EricWittmann: Same for all 3 readme files.

platforms/quarkus/api/README.md
EricWittmann: Same change needed for the -ws image as well.

carlesarnal: Everything should be ok with MySQL as well, just maybe some minor changes needed at config level. I'll check that and document how to configure the database type.

EricWittmann:
Thanks!

PR #1143, opened 2020-05-06
feat(CI/CD): Adds github actions workflows

riprasad:


riprasad:
@EricWittmann I tried using the same approach but it doesn't work for some reason. I get 401 error. As per my understanding, '.npmrc' file should be created at the root of the project. Since, there are two repositories being cloned here in two separate directories, the github action that allows us to create a '.npmrc, file has no way to specify the path where we want the file to be created. I also tried creating the file at the root of the directory manually, but I continue to get 401 error.

I tried both the approaches and none worked. Therefore, I had to use this approach.

PR #1188, opened 2020-06-19
Fix download servlet for quarkus deployment

carlesarnal:


front-end/servlet/src/main/java/io/apicurio/studio/fe/servlet/servlets/JwtDownloadServlet.java
```diff
@@ -0,0 +1,209 @@
...
+import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+import java.util.Date;
+
+/**
+ * Servlet used to download content (e.g. an API design).
+ * @author eric.wittmann@gmail.com
```
EricWittmann, line 52: I love when I get credit for stuff I didn't write.  I'm definitely going to show this to Paolo during my performance review.  ;)

front-end/servlet/src/main/java/io/apicurio/studio/fe/servlet/servlets/JwtDownloadServlet.java
```diff
@@ -0,0 +1,209 @@
...
+ */
+public class JwtDownloadServlet extends HttpServlet {
```
EricWittmann, line 54: Can we share a bunch of code between the JWT download servlet and the KC download servlet?

front-end/servlet/src/main/java/io/apicurio/studio/fe/servlet/servlets/JwtDownloadServlet.java
carlesarnal: Well, it’s quite the same content, so you can get all the credit :).

front-end/servlet/src/main/java/io/apicurio/studio/fe/servlet/servlets/JwtDownloadServlet.java
carlesarnal: Sure, I’ve tried to do as less changes as possible, but is the best to do.

PR #1213, opened 2020-07-15
Update StudioConfigServlet.java

nikolaistraessle:
Content type of the config.js should be "application/javascript" instead "application/json" to prevent strict content-type checking.

See https://en.it1352.com/article/d091144387194a23b319803ade7fa69c.html


EricWittmann:
Thanks!

PR #1236, opened 2020-08-05
Fix invalid default keycloak password in docker-compose readme

gsaslis:
This looks like it is probably outdated. 

The keycloak password is generated in setup.sh [here](https://github.com/Apicurio/apicurio-studio/blob/master/distro/docker-compose/setup.sh#L27) and used in docker-compose [here](https://github.com/Apicurio/apicurio-studio/blob/master/distro/docker-compose/docker-compose.keycloak.yml#L27) - via the .env file [here](https://github.com/Apicurio/apicurio-studio/blob/master/distro/docker-compose/.env.template#L17).


EricWittmann:
Thanks!

Issue #1250, opened 2020-08-21
Microcks Integration does not work with URL starting with https://api


franznemeth:
We have a kubernetes Setup of apicurio-studio with microcks integration.

Mocking an API with microcks works but the display URL in apicurio-studio is incorrect if the microcks URL starts with https://api.

For example: 
The endpoint for our apicurio-studio is apicurio-studio.example.org and the endpoint for microcks is apicurio-microcks.example.org

This results in apicurio writing the following mockURL to the database: "https://#/services/someapi%3A1.0.0"

The problem is this line of code.
https://github.com/Apicurio/apicurio-studio/blob/d038841eb5459062716616f65711cc86c4650b56/back-end/hub-api/src/main/java/io/apicurio/hub/api/rest/impl/DesignsResource.java#L897

An easy fix would be to change the .indexOf() to .lastIndexOf().

EricWittmann:
Well that's a great bug report.  With a fix and everything.  You don't see that every day. :)

Since you have a fix for this, would you like to open a PR for it?  (no problem if not!)

franznemeth:
Done, PR is open

PR #1255, opened 2020-08-27
fixed microcks url starting with https://api

franznemeth:
This PR addresses #1250

This ensures that microcks URLs starting with https://api... are not parsed incorrectly


EricWittmann:
Thanks!

PR #1258, opened 2020-08-31
added APICURIO_SHARE_FOR_EVERYONE to WS deployment in k8s and helm

franznemeth:
The kubernetes and helm example deployments in distro have the same issue as #954. 

This PR adds the env var to the helm chart and the kubernetes yamls so the share for everyone feature works.


EricWittmann:
Thanks.

PR #1276, opened 2020-09-19
add checking if GenerationConfig.isLongInteger is true when generating JavaInterface

joongs4:
I added checking if GenerationConfig.isLongInteger is true when generating JavaInterface. if it doesn't check, JavaBean can have long typed fields but the interface take it as integer.

### DIFF
--- a/back-end/hub-codegen/src/test/java/io/apicurio/hub/api/codegen/OpenApi2QuarkusTest.java
+++ b/back-end/hub-codegen/src/test/java/io/apicurio/hub/api/codegen/OpenApi2QuarkusTest.java
@@ -98,6 +98,9 @@ public class OpenApi2QuarkusTest {
                         System.out.println(actual);
                         System.out.println("-----");
                     }
+                    if(name.contains("OrgsResource")){
+                        System.out.println("aa");
+                    }
                     Assert.assertEquals("Expected vs. actual failed for entry: " + name, normalizeString(expected), normalizeString(actual));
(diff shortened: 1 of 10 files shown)


joongs4:
plus
1. add missing dependency
2. fix exception when generating Zip file at OpenApi2JaxRsTest

back-end/hub-codegen/src/test/java/io/apicurio/hub/api/codegen/OpenApi2QuarkusTest.java
```diff
@@ -98,6 +98,9 @@ private void doFullTest(String apiDef, boolean updateOnly, String expectedFilesP
                         System.out.println(actual);
                         System.out.println("-----");
                     }
+                    if(name.contains("OrgsResource")){
+                        System.out.println("aa");
+                    }
```
EricWittmann, line 103: I'm guessing this was included by mistake.  :)

EricWittmann (changes requested):
Just remove the `println()` and this looks good to me.  Thanks!

back-end/hub-codegen/src/test/java/io/apicurio/hub/api/codegen/OpenApi2QuarkusTest.java
joongs4: Thanks ! I removed and pushed it :D

Issue #447, opened 2018-09-25
Able to support AsyncAPI standard in editor


dmvolod:
As AsyncAPI has it's own [AsyncAPI editor](https://github.com/asyncapi/editor) it would be better to join this format and visual support (AsyncAPI editor doesn't have this option and only yaml description -> visual supports).
@EricWittmann what do you think about it? Or this is crazy idea?

EricWittmann:
I'd love to see an Apicurio visual editor for AsyncAPI.  Right now this project doesn't have the engineering bandwidth for such an effort. :(

dmvolod:
@EricWittmann thanks for response. I will try to find some time for work on this feature if it will not be so difficult...

EricWittmann:
Well I think adding support for AsyncAPI is a huge amount of work.  The first task would probably be to implement a parser/model layer for it.  For OAI I've done that here:  

https://github.com/apicurio/oai-ts-core

So that's probably a good place to start - perhaps a new repository like `asyncapi-ts-core`.  

If you're up for that sort of thing that'd be great!  I'd need to start thinking about the best way to get asyncapi support into the rest of Apicurio.  There are a number of considerations across the various UI pages.

EricWittmann:
For what it's worth, AsyncAPI is starting to pick up some additional steam.  So providing support for it in Apicurio is climbing up the priority list. :)  No ETA or anything like that YET - but hopefully we can start making some progress in this area.

dmvolod:
Good to hear this new, @EricWittmann , thanks. I would like to participate, if will find enough time.

fmvilas:
FWIW, we now have a [parser for AsyncAPI 2.0.0](https://github.com/asyncapi/parser-js). We've been discussing on migrating to Typescript. It'd be great to have a list of tasks of what needs to be done in order to have AsyncAPI in Apicurio. We may be able to help :)

EricWittmann:
That's interesting, @fmvilas  - thanks for the pointer.  We've actually built our own unified data model implementation for AsyncAPI and OpenAPI here:  https://github.com/Apicurio/apicurio-data-models

That library is written in Java and is transpiled into Typescript so we can use it both in our back-end and also directly in our UI.

[...]

PR #1285, opened 2020-09-24
Adding channel and data types management, properties edition and cont…

lbroudoux:
…ext-menus on AsyncAPI editor

As discussed into #447, this is a second PR to pursue the AsyncAPI graphical editor into Apicurio!

This second PR embeds following changes and capabilities:
* Graphical `Channel` creation / deletion as well as support for context-menu,
* Support of `SchemaDefinition` management: creation, cloning and deletion,
* Schema properties graphical edition.

Some screenshots below:
Capture d’écran 2020-09-24 à 14 12 51
Capture d’écran 2020-09-24 à 14 13 05
Capture d’écran 2020-09-24 à 14 13 31


EricWittmann:
Very nice, thanks.  We'll need to do a release of data-models before this PR will pass.

lbroudoux:
Great! Meanwhile I'll do some changes to activate the "Mock with Microcks" button for AsyncAPI too.

EricWittmann:
Release is done:  https://github.com/Apicurio/apicurio-data-models/releases/tag/1.1.5

You should be able to update both the `pom.xml` and `package.json` with the new version and the CI build should then hopefully work.

lbroudoux:
Hey @EricWittmann ,

Just changing the `pom.xml` and `package.json` is not enough as you have made some other refactoring ;-)

The moves of `Aai20SchemaDefinition` to `AaiSchema` are causing some blocking issues. For instance, because `AaiComponents` having a `Map<String, AaiSchema> schemas`, we studio can no longer retrieve a data type definition by name, the `name` property still being hold by `Aai20SchemaDefinition` ...

To be analog to what we're having on OpenAPI, I think we should introduce an `AaiSchemaDefinition`, make use of this one in `AaComponents` and having the concrete `Aai20SchemaDefinition` inherit this new class.

EricWittmann:
OK sorry for that.  The schemas in components should all actually be `Aai20SchemaDefinition` instances, though.  Can you not just cast them?

lbroudoux:
Oh well yes of course, it should work ... but I was wondering why we would not reproduce same design than in `OAS30Components` ?

lbroudoux:
Hooray ! Build is OK ;-)

PR #1293, opened 2020-10-02
Fixing roadmap page link

Pierre-G:
On the Dashboard, "View our Roadmap" was pointing to http://www.apicur.io/roadmap/ that has moved to /studio/roadmap.
I also changed http to https on the two other links.


EricWittmann (approved):
Nice, thank you.

Issue #1113, opened 2020-04-13
Error viewing generated documentation when importing data types


EricWittmann:
From @Brijesh-briju here: https://github.com/Apicurio/apicurio-studio/issues/1049#issuecomment-612844575

I'm able to import datatype from another API but when i click Live documentation ..its throwing error
as given below------------------

```
Something went wrong...
Error downloading apicurio:2 Failed to fetch
Stack trace
Error: Error downloading apicurio:2 
Failed to fetch
    at https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:63:6628
    at https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:122:36080

TypeError: Failed to fetch
    at e.exports.f._onFinish (https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:122:38159)
    at e.exports.<anonymous> (https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:122:37076)
    at e.exports.s.emit (https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:75:19861)
    at _ (https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:75:26498)
    at https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:75:29476
    at e.exports.v.end (https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:75:29557)
    at e.exports.f.end (https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:122:40131)
    at Object.l.get (https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:63:71507)
    at https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js:122:35192
    at new Promise (<anonymous>)

ReDoc Version: 2.0.0-rc.26
Commit: 5bd2e62 
```

PR #1297, opened 2020-10-06
Dereference when generating documentation to support imported data types through preview documentation action

binsentsu:
Hi,

Dereferencing the api seems to fix issue #1113 for me.
Please share your thoughts if this fix is adequate.


EricWittmann:
Great change, actually.  Thank you.

Issue #1319, opened 2020-11-07
Microcks realm is not uptodate for their latest release - Causes 403 when listing APIs in Microcks


narioinc:
Environment - Ubuntu 20.04
APICurio release - BETA 2.465
Reproducible - Yes
Affected artefact - Microcks realm config JSON file
Potential fix available - Yes
Fix tested on - Ubuntu 20.04

Issue description:
Team, first my appreciation to the team on creating such a wonderful tool for microservices devs like me. It has become a indispensable tool for distributed teams in collaborating on microservcie development.
The BETA 2.46 release of APicurio has a small issue with the Microcks real config. The Microcks latest release requires a "user" role in the "microcks-app" client which is currently missing in the config that ships with BETA 2.46. This creates an issue where when we try to push the API from APiCurio to Microcks, the number of APIs shows up correctly in the dashboard but Microcks does not show any API list in the " API and Services " section in the UI. This is because of the missing "user" role which results in the API call throwign A 403

Microcks has their realm sample updated here: [https://github.com/microcks/microcks/blob/master/install/keycloak-microcks-realm-addons.json](https://github.com/microcks/microcks/blob/master/install/keycloak-microcks-realm-addons.json)

I was able to test this confi json by replacing the one in the current release and then going through the APICUrio installation steps and the issue gets rectified. 

Can the team please update the realm file based on the need of the Microcks latest release. If possible. Can I request a PR for the fix

EricWittmann:
@lbroudoux is this the file that needs updating?  https://github.com/Apicurio/apicurio-studio/blob/master/distro/docker-compose/config/keycloak/microcks-realm.json.template

narioinc:
Hi Eric, good day. Just a quick confirmation from my side. The file you mentioned is the one that needs the fix. You are correct. But I'll let @lbroudoux also confirm. 
Thanks.

lbroudoux:
Hi !

Yes that's correct! With Microcks `1.0.0` we have moved the role definitions from the realm scope to the `microcks-app` client scope so that it is easier to plug Microcks on an existing realm (no more role names collision).

Thanks Eric and Naresh!

narioinc:
I have pushed a  fix for this via #1323 . Please review and let me know.

PR #1323, opened 2020-11-16
Fix for Microcks template Keycloak realm file to support Microcks 1.0.0 onwards

narioinc:
@EricWittmann 

I am creating this pull request for the fix to the Microcks realm template file to as to fix the bug #1319. The Fix proposes that with Microcks 1.0.0, role definitions have moved from the realm scope to the microcks-app client scope so that it is easier to plug Microcks on an existing realm.

**PR tested on the following environment:**
Environment - Ubuntu 20.04
APICurio release - BETA 2.465
Affected artefact - Microcks realm config JSON file
Potential fix available - Yes
Fix tested on - Ubuntu 20.04

@lbroudoux would like to review this PR along with you to ensure correctness. Thanks


EricWittmann:
@lbroudoux Does this look good to you?

distro/docker-compose/config/keycloak/microcks-realm.json.template
```diff
@@ -297,8 +295,28 @@
...
       "security-admin-console": [],
+      "microcks-app": [
+        {
+          "id": "b70db266-0a25-48b8-a53f-5666febd06ea",
+          "name": "user",
+          "composite": false,
+          "clientRole": true,
+          "containerId": "61397ec1-d0fa-4b9b-a38a-c301326b7833",
+          "attributes": {}
+        }
```
lbroudoux, line 318: To be fully aligned, we should also add the `manager` and `admin` roles as in https://github.com/microcks/microcks/blob/master/install/docker-compose/keycloak-realm/microcks-realm-sample.json#L65

distro/docker-compose/config/keycloak/microcks-realm.json.template
```diff
@@ -845,6 +875,48 @@
...
+            "multivalued": "true",
+            "user.attribute": "foo",
+            "id.token.claim": "true",
+            "access.token.claim": "true",
+            "claim.name": "groups",
+            "jsonType.label": "String"
+          }
+        }
+      ]
+    },
```
lbroudoux, line 919: This is not related to a Microcks requirement

lbroudoux (changes requested):
The `user` roles is defined twice and the `admin` and `manager` roles are missing. Also I don't see where the `user` role is defined as a default for the `microcks-app` client.

As this template should be the result of a Keycloak export, it is quite hard to understand and diff because of all the ids and optional attributes. Wouldn't it be possible to use a simpler form like in https://github.com/microcks/microcks/blob/master/install/docker-compose/keycloak-realm/microcks-realm-sample.json ?  I'll try re-used an older Keycloak version to test things up becase we've switcher to v10.1 on the Microcks side.

narioinc:
@lbroudoux  @EricWittmann  I guess the template file can be changed to the version that is there in microcks github.
Since I saw that the current template file seems to have been created from an export with all the finer details, I too took did a export and sent in a PR. 
I will create a PR with just the realm config provided mi microcks team here
https://github.com/microcks/microcks/blob/master/install/docker-compose/keycloak-realm/microcks-realm-sample.json

narioinc:
I have changed the file to reflect the correct concise template needed for the Microcks app to run. Checked a quick install it works on my environment

EricWittmann (approved):
LGTM waiting for @lbroudoux to approve.

lbroudoux (approved):

lbroudoux:
LGTM too ! Thanks @narioinc for PR !

narioinc:
thanks @lbroudoux for the review and approval !

PR #1327, opened 2020-11-20
Create quarkus docker compose distro 1321

carlesarnal:


distro/quarkus-docker-compose/.env.template
```diff
@@ -0,0 +1,68 @@
...
+MR_KEYCLOAK_URL=http://$HOST:8090/auth
+
+#APICURIO-STUDIO
+
+AS_DATABASE=apicuriodb
+AS_MYSQL_ROOT_PASSWORD=$AS_MYSQL_ROOT_PASSWORD
+AS_DB_USER=apicurio
+AS_DB_PASSWORD=$AS_DB_PASSWORD
+
+APICURIO_KC_AUTH_URL=http://$HOST:8090/auth/realms/apicurio
```
famartinrh, line 42: I think this is incorrect, it should be APICURIO_KC_AUTH_URL=http://$HOST:8090/auth

carlesarnal: No, for the Quarkus app, the correct value for the auth URL has to be like `http://$HOST:$PORT/auth/realms/$REALM`.

EricWittmann:
Ideally I think we want to phase out the Thorntail based deployments.  In your view, what is the path to making that happen?  In particular, I'd simply like to replace the existing docker images with ones based on Quarkus.  So ideally all of the ENV VARs and everything should be the same - or rather *usage* of the docker image should be the same.

carlesarnal:
> Ideally I think we want to phase out the Thorntail based deployments. In your view, what is the path to making that happen? In particular, I'd simply like to replace the existing docker images with ones based on Quarkus. So ideally all of the ENV VARs and everything should be the same - or rather _usage_ of the docker image should be the same.

And that's the case, all the ENV VARs are shared. I'm creating this as an alternative to the already existing `docker-compose` deployment based on the Thorntail images. The main differences here are located in the docker-compose files. For what's relative to the ENV VARs, everything is still the same, except for some values for those ENV VARs, e.g. the `APICURIO_KC_AUTH_URL` env var has to contain the Realm value for the Quarkus deployment.

Maybe this is a good topic for a Friday tech discussion, isn't it?

carlesarnal:
> Ideally I think we want to phase out the Thorntail based deployments. In your view, what is the path to making that happen? In particular, I'd simply like to replace the existing docker images with ones based on Quarkus. So ideally all of the ENV VARs and everything should be the same - or rather _usage_ of the docker image should be the same.

Right now it should be possible to replace the Thorntail images and we should be ok except for some bugs like [this](https://github.com/Apicurio/apicurio-studio/issues/1328). That's why I'm creating a distro like this one, to test everything properly and give the users the exact same experience that they already have without any new bugs (or with a minimum impact).

EricWittmann:
OK sounds good to me.  Let me know if/when you want this merged, if that makes using it easier.  If not, happy to keep it in draft mode as long as you need it.

carlesarnal:
> OK sounds good to me. Let me know if/when you want this merged, if that makes using it easier. If not, happy to keep it in draft mode as long as you need it.

Ok, I'll do so. I'm happy in having this in a draft state since there are a few things that need to be addressed but this way everyone might be aware of the existence of this.

carlesarnal:
Closed in favour of #1341

PR #1333, opened 2020-11-25
integration events - initial implementation and http cloudevents

famartinrh:
Initial implementation of integration events to allow third party apps to easily integrate with apicurio-studio. This provides some interfaces in order to be able to implement different protocols for sending the events. For now only http cloudevents is implemented.

Given the kind of application apicurio-studio is, being able to configure the "firing" of this integrations events in the UI would make a lot of sense. However, because of simplicity, for now configuration is only implemented with env vars and system properties.

So, things remaining for future contributions are:
* configuration in the UI / persistence of the configurations in the DB
* more event protocols, such as kafka
* more event types, at first only design created, updated and deleted are available
* improvements in existing event protocol implementations, such as rate limiting, authentication, retries,...

### DIFF
--- a/back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/ApicurioEventType.java
+++ b/back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/ApicurioEventType.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2020 Red Hat
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.apicurio.hub.core.integrations;
+
+/**
+ * @author Fabian Martinez
+ */
+public enum ApicurioEventType {
+
+    DESIGN_CREATED,
+    DESIGN_UPDATED,
+    DESIGN_DELETED;
+
+    private String cloudEventType;
+
+    private ApicurioEventType() {
+        this.cloudEventType = "io.apicurio.hub."+this.name().toLowerCase().replace("_", "-");
+    }
+
+    public String cloudEventType() {
+        return this.cloudEventType;
+    }
+
+}
--- a/back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/ApicurioEventType.java
+++ b/back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/ApicurioEventType.java
@@ -20,14 +20,14 @@ package io.apicurio.hub.core.integrations;
  */
 public enum ApicurioEventType {
 
-    DESIGN_CREATED,
-    DESIGN_UPDATED,
-    DESIGN_DELETED;
+    DESIGN_CREATED("io.apicurio.hub.design-created"),
+    DESIGN_UPDATED("io.apicurio.hub.design-updated"),
+    DESIGN_DELETED("io.apicurio.hub.design-deleted");
 
     private String cloudEventType;
 
-    private ApicurioEventType() {
-        this.cloudEventType = "io.apicurio.hub."+this.name().toLowerCase().replace("_", "-");
+    private ApicurioEventType(String cloudEventType) {
+        this.cloudEventType = cloudEventType;
     }
 
     public String cloudEventType() {
--- a/back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/EventSink.java
+++ b/back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/EventSink.java
@@ -9,7 +9,7 @@ public interface EventSink {
 
     boolean isConfigured();
 
-    void handle(ApicurioEventType type, String designId, byte[] data);
+    void handle(ApicurioEventType type, byte[] data);
 
 }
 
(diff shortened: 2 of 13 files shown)


EricWittmann:
I will review after the holiday (next week).  Initial skim looks great, thanks!

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/ApicurioEventType.java
```diff
@@ -0,0 +1,37 @@
...
+public enum ApicurioEventType {
+
+    DESIGN_CREATED,
+    DESIGN_UPDATED,
+    DESIGN_DELETED;
+
+    private String cloudEventType;
+
+    private ApicurioEventType() {
+        this.cloudEventType = "io.apicurio.hub."+this.name().toLowerCase().replace("_", "-");
```
carlesarnal, line 30: Suggestion, maybe add a field to the enum and use that instead of replacing it this way?

carlesarnal (dismissed):
LGTM

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/ApicurioEventType.java
EricWittmann: I agree - that would be a nice improvement.

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/EventsServiceImpl.java
```diff
@@ -0,0 +1,106 @@
...
+               }
+            });
+        }
+    }
+
+    private synchronized ObjectMapper getMapper() {
+        if (mapper == null) {
+            mapper = new ObjectMapper();
+            mapper.setSerializationInclusion(Include.NON_NULL);
+        }
```
EricWittmann, line 102: Maybe initialize the mapper in the `init()` method to remove this synchronized method?  Or if you prefer the lazy init approach, I suggest you remove `synchronized` from the method.  Because who cares if you actually create more than 1 due to the race condition?  In the case of `ObjectMapper` that doesn't really matter.  It's not worth the cost of synchronization IMO.

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/EventSink.java
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * @author Fabian Martinez
+ */
+public interface EventSink {
+
+    String name();
+
+    boolean isConfigured();
+
+    void handle(ApicurioEventType type, String designId, byte[] data);
```
EricWittmann, line 12: I imagine we might eventually have events that are associated with an API design.  But that's ok - let's keep the `designId` param here for now and we can either modify or add another `handle()` variant in the future **if we need it**. :)

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/http/HttpEventSink.java
```diff
@@ -0,0 +1,99 @@
...
+        return "http sink";
+    }
+
+    @Override
+    public boolean isConfigured() {
+        return configuration.isConfigured();
+    }
+
+    @Override
+    public void handle(ApicurioEventType type, String designId, byte[] data) {
```
EricWittmann, line 56: The API design id is not used?

EricWittmann: Maybe I've changed my mind about the interface method. :)

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/http/HttpEventSink.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    private synchronized HttpClient getHttpClient() {
+
+        if (httpClient == null) {
+            httpClient = HttpClientBuilder.create().build();
+        }
+        return httpClient;
+    }
```
EricWittmann, line 97: Initialize in a `@PostConstruct` method to avoid the synchronized method?

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/ApicurioEventType.java
EricWittmann: Although thinking about it more, the result is pretty constrained - it's not like we don't have control of the result and may get something unexpected.  I'm 50/50 on this.  Whatever Fabian decides is fine. :)

EricWittmann:
@famartinrh Very nice - just a few comments for you to address before merging.

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/http/HttpEventSink.java
famartinrh: I think that would result in the http client always being initialized, even when not configuring it. Which is what I tried to avoid

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/http/HttpEventSink.java
famartinrh: hehe I forgot about this :), I'll remove that argument

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/EventsServiceImpl.java
famartinrh: ok, I'll remove the `synchronized` even knowing there is a possible race condition

back-end/hub-core/src/main/java/io/apicurio/hub/core/integrations/http/HttpEventSink.java
EricWittmann: You could create the client only if `configuration.isConfigured()` returns `true`, no?

EricWittmann:
```java
@PostConstruct
void init() {
    if (configuration.isConfigured()) {
        httpClient = HttpClientBuilder.create().build();
    }
}
```

EricWittmann:
Looks great, thanks @famartinrh !

PR #1370, opened 2021-01-07
Fix member access on undefined (ImportedComponent.from is optional)

DyspC:
Hello, 

I'm currently looking into ways to use an external component source (namely an apicurio-registry instance) in the studio import wizard and I've come across this inconsistency: the ImportedComponent.from field is optional and the review panel tries to read its name anyway, causing the wizard component to get stuck on the error report.

This PR wraps the "{ic.from} /" part into a conditional block to prevent this.

---
On a side note, I just found this issue and this is what I'm trying to add to a corporate POC
https://github.com/Apicurio/apicurio-registry/issues/752#issuecomment-688859362


EricWittmann (approved):

EricWittmann:
Thanks for the PR - it sounds like you're doing something very interesting.  I'd love to see what you come up with if you're willing to demo it at some point!

PR #1381, opened 2021-01-21
Add apicurio-registry as a possible external component source

DyspC:
Hi @EricWittmann, 

As exposed in #1370, I'm working on the possibility to use the registry as a component source, with the idea that we will normalize our openapi/asyncapi specifications by externalizing data types and have them being referenced at an immutable revision.

I'm sending you this first draft so we can discuss the implementation and ways it could be enhanced.

As you'll see I tried to fit the overall style and architecture of the project while laying ground for the backend to act as a proxy if the registry needs to be protected from the users.

I edited the import-component-wizard in a way it joins the results from the 'designs' and 'registry' endpoints in the same list and visits all resources when needed (only for ASYNCAPI, OPENAPI and JSON artifact types atm). The selected components are then added as $ref to the latest published component version.

image

I left some to-do tasks in the front end as I know very little of typescript and nothing of angular so I'd need someone's help on this part.

I also have performance concerns with the wizard fetching the entire registry inventory (does that happen with all APIs if shareWithEveryone is enabled?) but adding another wizard with search functionalities is over what I can do with angular.

If you want to run it you just have to add an APICURIO_REGISTRY_URL prop (from http to port number) to the api and ui modules.

### DIFF
--- a/back-end/hub-api/src/main/java/io/apicurio/hub/api/content/ContentDereferencer.java
+++ b/back-end/hub-api/src/main/java/io/apicurio/hub/api/content/ContentDereferencer.java
@@ -40,11 +40,14 @@ public class ContentDereferencer {
     private AbsoluteReferenceResolver absoluteResolver;
     @Inject
     private InternalReferenceResolver apicurioResolver;
+    @Inject
+    private RegistryReferenceResolver registryResolver;
 
     @PostConstruct
     void init() {
         Library.addReferenceResolver(apicurioResolver);
         Library.addReferenceResolver(absoluteResolver);
+        Library.addReferenceResolver(registryResolver);
     }
     
     /**
--- a/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/IRegistry.java
+++ b/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/IRegistry.java
@@ -6,10 +6,30 @@ import io.apicurio.registry.rest.beans.SearchedArtifact;
 import java.io.IOException;
 import java.util.Collection;
 
+/**
+ * @author c.desc2@gmail.com
+ */
 public interface IRegistry {
-    Collection<SearchedArtifact> listArtifacts() throws IOException;
 
+    /**
+     * Gets all the artifacts the configured registry instance can serve
+     * @return a {@link Collection} of all the {@link SearchedArtifact} objects
+     */
+    Collection<SearchedArtifact> listArtifacts();
+
+    /**
+     * Gets the string resource represented in the configured registry by the given artifactId and artifactVersion
+     * @param artifactId the artifactId
+     * @param artifactVersion the artifactVersion
+     * @return the artifact content
+     * @throws IOException if the content could not be read as UTF_8
+     */
     String getArtifact(String artifactId, Integer artifactVersion) throws IOException;
 
-    ArtifactMetaData getArtifactMetaData(String artifactId) throws IOException;
+    /**
+     * Gets the {@link ArtifactMetaData} for the given artifactId
+     * @param artifactId the artifactId
+     * @return the {@link ArtifactMetaData}
+     */
+    ArtifactMetaData getArtifactMetaData(String artifactId);
 }
--- a/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/IRegistry.java
+++ b/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/IRegistry.java
@@ -7,6 +7,8 @@ import java.io.IOException;
 import java.util.Collection;
 
 /**
+ * This interface defines the possible actions the Apicurio Studio can perform on a configured Apicurio Registry
+ * As of now, the Studio only gets artifacts and their related metadata / content
  * @author c.desc2@gmail.com
  */
 public interface IRegistry {
--- a/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/IRegistry.java
+++ b/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/IRegistry.java
@@ -1,0 +1,15 @@
+package io.apicurio.hub.core.registry;
+
+import io.apicurio.registry.rest.beans.ArtifactMetaData;
+import io.apicurio.registry.rest.beans.SearchedArtifact;
+
+import java.io.IOException;
+import java.util.Collection;
+
+public interface IRegistry {
+    Collection<SearchedArtifact> listArtifacts() throws IOException;
+
+    String getArtifact(String artifactId, Integer artifactVersion) throws IOException;
+
+    ArtifactMetaData getArtifactMetaData(String artifactId) throws IOException;
+}
--- a/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/http/HttpRegistry.java
+++ b/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/http/HttpRegistry.java
@@ -28,7 +28,7 @@ public class HttpRegistry implements IRegistry {
 
     @Inject
     public HttpRegistry(HubConfiguration config) {
-        this.registryRestClient = Optional.ofNullable(config.getRegistryUrl())
+        this.registryRestClient = Optional.ofNullable(config.getRegistryApiUrl())
                 .map(RegistryRestClientFactory::create)
                 .orElse(null);
     }
--- a/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/http/HttpRegistry.java
+++ b/back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/http/HttpRegistry.java
@@ -1,0 +1,99 @@
+package io.apicurio.hub.core.registry.http;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.apicurio.hub.core.config.HubConfiguration;
+import io.apicurio.hub.core.registry.IRegistry;
+import io.apicurio.registry.rest.beans.ArtifactMetaData;
+import io.apicurio.registry.rest.beans.ArtifactSearchResults;
+import io.apicurio.registry.rest.beans.SearchedArtifact;
+import org.apache.commons.io.IOUtils;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.entity.ContentType;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Default;
+import javax.inject.Inject;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+@ApplicationScoped
+@Default
+public class HttpRegistry implements IRegistry {
+
+    private static final Logger logger = LoggerFactory.getLogger(HttpRegistry.class);
+    private static CloseableHttpClient httpClient = HttpClients.createDefault();
+    private static final ObjectMapper mapper = new ObjectMapper();
+
+    @Inject
+    private HubConfiguration config;
+    
+    @Override
+    public Collection<SearchedArtifact> listArtifacts() throws IOException {
+        List<SearchedArtifact> artifactMetaDataList = Collections.emptyList();
+        if (config.getRegistryUrl() != null) {
+            HttpGet get = new HttpGet(String.format("%s/api/search/artifacts", config.getRegistryUrl()));
+            try (CloseableHttpResponse response = httpClient.execute(get)) {
+                if (response.getStatusLine().getStatusCode() >= 200 && response.getStatusLine().getStatusCode() < 300) {
+                    try (InputStream contentStream = response.getEntity().getContent()) {
+                        final ArtifactSearchResults artifactSearchResults = mapper.readValue(contentStream, ArtifactSearchResults.class);
+                        artifactMetaDataList = artifactSearchResults.getArtifacts();
+                    }
+                }
+            }
+        }
+        return artifactMetaDataList;
+    }
+
+    @Override
+    public String getArtifact(String artifactId, Integer artifactVersion) throws IOException {
+        String content = "";
+        if (config.getRegistryUrl() != null) {
+            StringBuilder urlBuilder = new StringBuilder();
+            urlBuilder.append(config.getRegistryUrl())
+                    .append("/api/artifacts/").append(artifactId);
+            if (artifactVersion != null) {
+                urlBuilder.append("/versions/").append(artifactVersion);
+            }
+            HttpGet get = new HttpGet(urlBuilder.toString());
+            try (CloseableHttpResponse response = httpClient.execute(get)) {
+                if (response.getStatusLine().getStatusCode() >= 200 && response.getStatusLine().getStatusCode() < 300) {
+                    try (InputStream contentStream = response.getEntity().getContent()) {
+                        ContentType ct = ContentType.getOrDefault(response.getEntity());
+                        String encoding = StandardCharsets.UTF_8.name();
+                        if (ct != null && ct.getCharset() != null) {
+                            encoding = ct.getCharset().name();
+                        }
+                        content = IOUtils.toString(contentStream, encoding);
+                    }
+                }
+            }
+        }
+        return content;
+    }
+
+    @Override
+    public ArtifactMetaData getArtifactMetaData(String artifactId) throws IOException {
+        ArtifactMetaData metadata = null;
+        if (config.getRegistryUrl() != null) {
+            HttpGet get = new HttpGet(String.format("%s/api/artifacts/%s/meta", config.getRegistryUrl(), artifactId));
+            try (CloseableHttpResponse response = httpClient.execute(get)) {
+                if (response.getStatusLine().getStatusCode() >= 200 && response.getStatusLine().getStatusCode() < 300) {
+                    try (InputStream contentStream = response.getEntity().getContent()) {
+                        metadata = mapper.readValue(contentStream, ArtifactMetaData.class);
+                    }
+                }
+            }
+        }
+        return metadata;
+    }
+
+}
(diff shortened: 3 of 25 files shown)


EricWittmann:
This is really interesting!  I'll need to spend some brain power thinking about this more than I have time for right now.  However, I did skim through the changes a little bit and have one mechanical suggestion for you.  It would probably be better and easier to use the Apicurio Registry REST Client in the implementation of `IRegistry` rather than a plain HTTP client.

https://github.com/Apicurio/apicurio-registry/blob/master/rest-client/src/main/java/io/apicurio/registry/client/RegistryRestClientFactory.java

Regarding the performance concern - I agree that we will need to make sure the wizard implements paging for all resources we display there.  Paging and filtering will both be important.  I can't recall what the implementation is right now, but I'm guessing it's fetching everything and filtering/ordering in the browser.  That's a poor long term solution if that's what is happening.

back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/http/HttpRegistry.java
```diff
@@ -24,74 +18,49 @@
...
-    private static CloseableHttpClient httpClient = HttpClients.createDefault();
-    private static final ObjectMapper mapper = new ObjectMapper();
+    private final RegistryRestClient registryRestClient;
 
     @Inject
-    private HubConfiguration config;
-    
+    public HttpRegistry(HubConfiguration config) {
+        this.registryRestClient = Optional.ofNullable(config.getRegistryUrl())
+                .map(url -> url.endsWith("/") ? url + "api" : url + "/api")
```
EricWittmann, line 32: I think this should be removed, and anyone configuring the system should provide the full endpoint to the root of the API.  The reason is that the next version of Apicurio Registry will be moving around the endpoints a bit, due to a forthcoming "v2" REST API.  So there will be two endpoints:  `/api/v1` and `/api/v2` - with that change it will become impossible to properly configure the endpoint here.

back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/IRegistry.java
```diff
@@ -0,0 +1,15 @@
+package io.apicurio.hub.core.registry;
+
+import io.apicurio.registry.rest.beans.ArtifactMetaData;
+import io.apicurio.registry.rest.beans.SearchedArtifact;
+
+import java.io.IOException;
+import java.util.Collection;
+
+public interface IRegistry {
```
EricWittmann, line 9: Some javadoc on this would be good.  I'm assuming this interface is how Studio interacts with an instance of Registry?  I wonder - do you anticipate multiple implementations of this?  Currently the only way for Studio to access Registry is via its REST API.

front-end/studio/src/app/pages/apis/{apiId}/_components/import-components.wizard.ts
```diff
@@ -53,8 +56,8 @@ export class ImportComponentsWizard {
     type: ComponentType;
 
     private result: Promise<ImportedComponent[]>;
-    private allApis: Api[];
-    selectedApis: Api[];
+    private allResources: (Api | Artifact)[]; // TODO are type unions a bad way to handle this?
+    selectedResources: (Api | Artifact)[];
```
EricWittmann, line 60: No - I like type unions. :)

front-end/studio/src/app/pages/apis/{apiId}/_components/import-components.wizard.ts
```diff
@@ -289,28 +304,37 @@ export class ImportComponentsWizard {
...
-        console.debug("[ImportComponentsWizard] Loading components from %o resources.", this.selectedApis.length);
-        let promises: Promise<ApiDefinition>[] = this.selectedApis.map( api => {
-            console.debug("[ImportComponentsWizard] Loading api definition for: ", api.name);
-            return this.apis.getApiDefinition(api.id).then(apiDef => {
-                this.processLoadedApiDef(apiDef);
-                return apiDef;
-            });
+        console.debug("[ImportComponentsWizard] Loading components from %o resources.", this.selectedResources.length);
+        let promises: Promise<ApiDefinition | ArtifactDefinition>[] = this.selectedResources.map(apiOrArtifact => {
+            // TODO find a better way to route the processing flow (instanceof does not work)
```
EricWittmann, line 324: Agreed - probably need to annotate the apis and artifacts with some sort of routing property.  Something like `__resourceType` which would be either `API` or `ARTIFACT`.

EricWittmann: This can be done in their respective Promises perhaps?

front-end/studio/src/app/services/artifacts.service.ts
```diff
@@ -0,0 +1,101 @@
...
+     */
+    constructor(http: HttpClient, authService: IAuthenticationService, config: ConfigService) {
+        super(http, authService, config);
+    }
+
+    /**
+     * @see ArtifactsService.getArtifacts
+     */
+    public getArtifacts(): Promise<Artifact[]> {
+        console.info("[ApisService] Getting all Artifacts");
```
EricWittmann, line 50: Copy/paste error - should be `ArtifactsService`.

shared/config/src/main/java/io/apicurio/studio/shared/config/Configuration.java
```diff
@@ -35,6 +35,9 @@
     private static final String KC_AUTH_REALM_ENV = "APICURIO_KC_AUTH_REALM";
     private static final String KC_AUTH_REALM_SYSPROP = "apicurio.kc.auth.realm";
 
+    private static final String REGISTRY_URL_ENV = "APICURIO_REGISTRY_URL";
+    private static final String REGISTRY_URL_SYSPROP = "apicurio-registry.url";
```
EricWittmann, line 39: Can you change this to `apicurio.registry.url`?

EricWittmann:
OK I have some minor suggestions for changes, but overall this is really good I think.

Looking at the result of the document, I think we should modify the URL being inserted as the `$ref`.  I think instead of an absolute URL getting inserted for the artifact, instead we should use a format similar to what we are currently doing for references to other Studio APIs.  For example, an imported type in Studio to another Studio API might be this:

`apicurio:40941#/components/schemas/ErrorInfo`

I suggest something similar for this use-case.  Maybe something like this:

`apicurio-registry:MyRefClient/1#/components/schemas/FooType`

The URL format would then be `apicurio-registry:<ARTIFACT_ID>/<VERSION>#<PATH>`

We can then enhance our reference resolvers to resolve this format in addition to the current `apicurio` internal one.

That would remove the need for the registry URL to be pushed into the UI, and it would make it possible to move the registry to a new location without breaking content.  

Thoughts?

back-end/hub-core/src/main/java/io/apicurio/hub/core/registry/IRegistry.java
DyspC: I didn't know the REST API was the only way through, but coming from Spring I would have defined a no-op implementation for when the registry URL is not set. This could still be used for a MockRegistry implementation if necessary.

EricWittmann: Cool.  I don't mind the extra layer of abstraction.  It's probably not going to be needed, but it'll be nice to have if I'm wrong.

back-end/hub-api/src/main/java/io/apicurio/hub/api/content/ContentDereferencer.java
```diff
@@ -40,11 +40,14 @@
...
     @Inject
     private InternalReferenceResolver apicurioResolver;
+    @Inject
+    private RegistryReferenceResolver registryResolver;
 
     @PostConstruct
     void init() {
         Library.addReferenceResolver(apicurioResolver);
         Library.addReferenceResolver(absoluteResolver);
+        Library.addReferenceResolver(registryResolver);
```
EricWittmann, line 50: Perfect.

EricWittmann:
I forgot about that link as well.  Maybe keep a URL to service registry in the UI config, but it should be the link to the registry UI (as you suggest) instead of the API URL.  That would rock, actually.  Having an external link to the registry UI from here would be super cool.

As for the javadoc, I was actually just suggesting class level documentation to add a quick explanation about what the interface was used for. :)  The methods are probably more self explanatory once you know what the interface is for.

DyspC:
I did not dive into the registry that much but is anyone with access to the registry (either UI or API) able to create new versions / delete artifacts?
This is why I initially used the hub API as a proxy instead of having the front-end call the registry; that way the registry could be on a private network and nobody would be able to erase referenced content.
Am I missing something?

Anyway, the former registry URL property is now split into UI and API (resp. for the UI and API modules of the studio), 
- if the registry API is missing new imports will only use the internal storage;
- if the registry UI is missing the definition-from component will not render a link at all, but the component preview will still be present;

image

EricWittmann:
> I did not dive into the registry that much but is anyone with access to the registry (either UI or API) able to create new versions / delete artifacts?
> This is why I initially used the hub API as a proxy instead of having the front-end call the registry; that way the registry could be on a private network and nobody would be able to erase referenced content.
> Am I missing something?

The current released version of the registry does not have authn/z support, but that was always just temporary.  Version 2.0 (the next release) will have authn/z and so we will need to be able to configure auth credentials for this as well.  Ideally we would support both BASIC and OAuth-ClientCredentials flows.  But that doesn't' need to happen in this PR.  That's a future change.

I am 100% in favor of using the studio API as a proxy to the registry.  If I said something to the contrary I was being unclear.  :)

EricWittmann:
> Anyway, the former registry URL property is now split into UI and API (resp. for the UI and API modules of the studio),
> 
>     * if the registry API is missing new imports will only use the internal storage;
> 
>     * if the registry UI is missing the definition-from component will not render a link at all, but the component preview will still be present;

Perfect!  Looks awesome.  :)

I think I was suggesting that we didn't need the registry API URL pushed into the UI layer because all querying and fetching of registry data would go through the proxy, and the UI already knows where the hub API lives.  We should only need the push the Registry UI URL into the Studio UI config, right?

I don't think we would use the Registry API URL for anything in the Studio UI layer....

DyspC:
> > Anyway, the former registry URL property is now split into UI and API (resp. for the UI and API modules of the studio),
> > ```
> > * if the registry API is missing new imports will only use the internal storage;
> > 
> > * if the registry UI is missing the definition-from component will not render a link at all, but the component preview will still be present;
> > ```
> 
> Perfect! Looks awesome. :)
> 
> I think I was suggesting that we didn't need the registry API URL pushed into the UI layer because all querying and fetching of registry data would go through the proxy, and the UI already knows where the hub API lives. We should only need the push the Registry UI URL into the Studio UI config, right?
> 
> I don't think we would use the Registry API URL for anything in the Studio UI layer....

I think this is what I changed when I split the properties?

The old property was defined in all Configuration instances and now the registry API URL is in HubConfiguration (https://github.com/Apicurio/apicurio-studio/pull/1381/commits/b3e17871d50a1d0387d718cbcb582497b94e3b7c#diff-7fdc5bb119f66ae98d85aad4db53fc3907e4a3e4f82466dff24139abbcaabd54R109) and the registry UI URL is in StudioUiConfiguration (https://github.com/Apicurio/apicurio-studio/pull/1381/commits/b3e17871d50a1d0387d718cbcb582497b94e3b7c#diff-a9e3dc2fe5dcc3cd903b17c8d186703f21a819d78324e06706cacd775772249fR57).
Then only the UI URL is sent to the front-end in the StudioConfigServlet (https://github.com/Apicurio/apicurio-studio/pull/1381/commits/b3e17871d50a1d0387d718cbcb582497b94e3b7c#diff-c8ff29217bb8d230045212a963d3c7ca64a45001dc6b8dc3c12c1a52762596c3R115)

EricWittmann:
OK awesome - we're on the same page. :)

DyspC:
> > I did not dive into the registry that much but is anyone with access to the registry (either UI or API) able to create new versions / delete artifacts?
> > This is why I initially used the hub API as a proxy instead of having the front-end call the registry; that way the registry could be on a private network and nobody would be able to erase referenced content.
> > Am I missing something?
> 
> The current released version of the registry does not have authn/z support, but that was always just temporary. Version 2.0 (the next release) will have authn/z and so we will need to be able to configure auth credentials for this as well. Ideally we would support both BASIC and OAuth-ClientCredentials flows. But that doesn't' need to happen in this PR. That's a future change.

That's all good for us if it is on the roadmap, is the registry one accessible somewhere? I only found it for the studio there https://www.apicur.io/studio/roadmap/

> I am 100% in favor of using the studio API as a proxy to the registry. If I said something to the contrary I was being unclear. :)

There was no misleading, I just realized I never stated why I wanted a proxy rather than direct requests.

EricWittmann:
The closest I have right now to a roadmap for Registry is here:

https://github.com/Apicurio/apicurio-registry/projects/2

Shifting priorities has made roadmaps difficult to keep updated and accurate. :)

EricWittmann:
What is missing from this before it's ready for final review and merge? :)

DyspC:
Erm... 

The two registry dependencies aren't set to the same version, that caused deserialization issues when artifact dates were changed from longs to date formats and I updated my registry. I've not pushed this commit to the public branch yet. 

There is also an issue with the reference details (https://github.com/Apicurio/apicurio-studio/pull/1381#issuecomment-769913756) not being rendered on our corporate deployment, I should have time to look into this tomorrow and I think that will be all.

DyspC:
> The two registry dependencies aren't set to the same version, that caused deserialization issues when artifact dates were changed from longs to date formats and I updated my registry. I've not pushed this commit to the public branch yet.

Fixed

> There is also an issue with the reference details ([#1381 (comment)](https://github.com/Apicurio/apicurio-studio/pull/1381#issuecomment-769913756)) not being rendered on our corporate deployment, I should have time to look into this tomorrow and I think that will be all.

I think this is an issue with the aaieditor not using the ApiCatalogService, if I load my imports from an openapi editor beforehand those are cached and rendered correctly

EricWittmann:
> The two registry dependencies aren't set to the same version, that caused deserialization issues when artifact dates were changed from longs to date formats and I updated my registry. I've not pushed this commit to the public branch yet.

We're actually reverting that change in the "V1" registry API soon.  We're trying to maintain backwards compatibility in the registry by freezing the old API as V1 and introducing the new API as V2.  The V2 API will have proper ISO 8601 dates but we'll keep the V1 API using the old long format.  That was always a bug but it is a breaking change so it will have to stay.  

FYI :)

EricWittmann:
> > There is also an issue with the reference details ([#1381 (comment)](https://github.com/Apicurio/apicurio-studio/pull/1381#issuecomment-769913756)) not being rendered on our corporate deployment, I should have time to look into this tomorrow and I think that will be all.
> 
> I think this is an issue with the aaieditor not using the ApiCatalogService, if I load my imports from an openapi editor beforehand those are cached and rendered correctly

This should be easy to fix I believe.  @lbroudoux what do you think?  I can help out here probably.

lbroudoux:
Hey there! Great enhancement! I put a +100 on it 😉. Back from some vacations and will have a look at the issue with aaieditor.

lbroudoux:
Will it possible to also use that for referencing Avro schema from registry? I cannot see any difference with exposed examples but just asking...

DyspC:
> Will it possible to also use that for referencing Avro schema from registry? I cannot see any difference with exposed examples but just asking...

I think there would be no technical issue; I'd just have to whitelist the AVRO type [here](https://github.com/Apicurio/apicurio-studio/pull/1381/files#diff-a94dfcbb5887fb179fc6dc2875c5545f4c42452a246034136a829fc7e68d383bR367) and I think it would fit in the current JSON flow.

On the functional side though the [AsyncAPI specification](https://www.asyncapi.com/docs/specifications/2.0.0#properties) states the schema is a JSON Schema and it looks like at least the `type` field in avro having the 'record' value does not fit the [JSON Schema RFC](https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.1.1)

I don't know about the other supported document types in the studio, there are probably some that would allow such schemas.

lbroudoux:
> On the functional side though the AsyncAPI specification states the schema is a JSON Schema and it looks like at least the type field in avro having the 'record' value does not fit the JSON Schema RFC

You're right! So refining my previous question: in the case of AsyncAPI it would be nice to be able to also reference Avro schema from the `payload` property of message that is of `any` type. Typically the kind of construct below is valid and provided as example within the spec:

```json
  "payload": {
    "$ref": "https://apicurio-registry/user-create.avsc/1#/UserCreate"
  }
```

What do you think?

DyspC:
If I'm not mistaken the message-section component is missing the part where we edit the payload so I can't really picture how to handle this but I'd say that the import-components wizard currently only has the ComponentType to filter the eligible components.

Using the AsyncAPI example

This means we have to segregate `/components/messages/_/payload` and `/components/schemas/_` imports by either adding a new ComponentType like ComponentType.messagePayload or something like ComponentTarget (default 'document' or 'components')

Besides, your example ends with `#/UserCreate`, this means implementing an AvroComponentFinder and it goes a bit out of hand since I don't know anything about this format.

EricWittmann:
I really like this contribution - where have we landed on it?  :)

DyspC:
So far here is what I think was left pending

> I think this is an issue with the aaieditor not using the ApiCatalogService, if I load my imports from an openapi editor beforehand those are cached and rendered correctly

This is not related to the "registry as component source" functionality and is probably just a matter of using the service the same way the editor component does. I'll try adding that to my soon-to-be AAI Servers PR.

> Will it possible to also use that for referencing Avro schema from registry? I cannot see any difference with exposed examples but just asking...

As long as /components/schemas don't accept avro this will need the import to be reworked in a way that defines the import target instead of sending it all to /components/schemas and we don't have any internal usecase for asyncapi+avro ATM so I doubt I would get time for this.

EricWittmann:
> > Will it possible to also use that for referencing Avro schema from registry? I cannot see any difference with exposed examples but just asking...
> 
> As long as /components/schemas don't accept avro this will need the import to be reworked in a way that defines the import target instead of sending it all to /components/schemas and we don't have any internal usecase for asyncapi+avro ATM so I doubt I would get time for this.

No problem.  I don't think there should be any reason to not accept this contribution just because it does not work for aai+avro.  Let's make sure it works well for the use-cases for which it is designed, and then go from there. :)

EricWittmann:
Closing as stale.  Sorry about this one.  I do think that the path forwards is external integration, rather than direct integration between the two.  Hopefully I can finally make some progress in this area soon.

PR #1384, opened 2021-01-22
fix: Improvements and fixes to helm chart

willemm:
This pull request represents the changes I had to make to the helm chart to get it working on our cluster.

So I'm making this PR to the upstream in the hopes others find it useful.

(After getting it working, I took all specific values and moved them into the values-yaml so the chart is more generic.)

Changes I did:
- Add `--ignore-db-dir=lost+found` to the mysql arguments, because some persistent volumes come with a lost+found directory
- Add `securityContext: fsGroup: 999` to the mysql deployment because otherwise mysql doesn't have write permissions on its storage volume
- Enable the option of adding extra volumemounts to pods (for us this was needed to mount the public-rootca of our company proxy)
- Change the ingress spec to be able to pass the hostname it's running on from the values (and use that hostname to point to the api and ws endpoints)
- Fixed a bug in the ws deployment where the database-connection was pointing to keycloak (a typo I guess)

### DIFF
--- a/distro/helm/templates/NOTES.txt
+++ b/distro/helm/templates/NOTES.txt
@@ -7,7 +7,7 @@ To learn more about the release, try:
   $ helm status {{ .Release.Name }}
   $ helm get {{ .Release.Name }}
 
-Apicurio is available at {{ .Values.ui.hub.api.url }}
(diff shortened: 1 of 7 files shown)


EricWittmann:
Thanks!

Issue #22, opened 2016-12-08
Does apiman-studio need i18n


msavy:
- Do we need it? 
- What's the best approach for this app?
- Should we use the [Angular 2](https://angular.io/docs/ts/latest/cookbook/i18n.html) i18n toolset. It seems vastly better than Angular v1's stuff.

Yuutakasan:
I want translation function by all means. Japanization cooperation is possible.

EricWittmann:
Unfortunately right now we do not have i18n support in Apicurio.  I'm hopeful that a re-working of the UI using Patternfly 4 (a process that will hopefully be happening in 2019) will help with this.  But there's no timeframe yet on it.

deka:
Please add i18n support.

Yuutakasan:
@EricWittmann 
I am planning to make apicurio multilingual this year.
Do you consider recreating the view layer in the future when you use PatternFly4?

If I make a multilingualization of the current view layout this year and send a merge request, would you be able to commit it?

EricWittmann:
Hi @Yuutakasan - we are currently working on replacing the UI with a React/PatternFly 4 implementation.  That project is ongoing here:  https://github.com/apicurio/apicurio-studio-react-poc

I would definitely recommend looking there for how you could contribute i18n support rather than do anything in the current generation UI.

Yuutakasan:
@EricWittmann 
I was doing something useless if I was working on it. Thanks for the information.

EricWittmann:
@dlabaj Could you provide any insight/feedback on the best way @Yuutakasan could perhaps help with i18n support in the new Apicurio Studio React POC?

Yuutakasan:
@EricWittmann 
If I can release the internationalization program by November 2020, I think I can develop Japanese translation and internationalization and send a Pull Request from August.

EricWittmann:
A Japanese translation would be fantastic, of course!  Hopefully PF4 has good i18n support that we can leverage.  @dlabaj any thoughts on that?

Yuutakasan:
@EricWittmann 
If you can provide us with the language files, we can translate them for you, and once the project starts, we may be able to assign an engineer to develop modifications to the view logic. Unfortunately, even if the project doesn't start, I can help with the translation.

turutosiya:
I'm looking for i18n support in Apicurio front-end and finally found this issue.
But unfortunately, it seems that `apicurio-studio-react-poc` has no commit recently...
Is there any progress with this?

[...]

PR #1391, opened 2021-01-27
Sample i18n (Japanese)

turutosiya:
Hi @EricWittmann ,

[Based on your suggestion](https://github.com/Apicurio/apicurio-studio/issues/22#issuecomment-756958467), I just imeplemted i18n feature for studio and create a very simple & small PR to evaluate the way to implement i18n.

In this PR, only the labels in the side menu are localized:

Screen Shot 2021-01-27 at 16 37 05

You can test this by running studio with `--configuration=ja` option.

```
$ ng serve --port 8888 --aot --configuration=ja
```


msavy:
I'm impressed, this seems quite unobtrusive!

EricWittmann:
This looks good to me!  And I agree is very unobtrusive.

Some questions I have after seeing this:

* This requires a separate build of the UI for each locale, correct?  If so, are they all bundled and released together so that the locale can be selected at runtime?  Or is a separate binary release of Studio needed for each locale?
* It looks good for simple labels, how does it work for more complex strings that perhaps have dynamic information injected?
* Does the i18n support also work with attributes?  (for example the `title` attribute)

turutosiya:
> * This requires a separate build of the UI for each locale, correct?  If so, are they all bundled and released together so that the locale can be selected at runtime?  Or is a separate binary release of Studio needed for each locale?

In this case, I tried with AOT compile option. So, it requires a separate build of the UI for each locale.
And I think we can bundle all of the locales together by adjusting `outputPath` of `angular.json`.
( Switch `outputPath` for each locales.  )

Besides, there is a `JIT` option.
With that option, we can switch language at runtime.
But, it will impact current code base and also the performance. 

> * It looks good for simple labels, how does it work for more complex strings that perhaps have dynamic information injected?

Let me try.

> * Does the i18n support also work with attributes?  (for example the `title` attribute)

Yes, of course. For attribute, we can do this way:

```
<img [src]="logo" i18n-title title="i18ned title" />
```

Also, we can translate without any tag :

```
<ng-container i18n>I don't output any element</ng-container>
```

EricWittmann:
OK thanks for answering my questions.  I think this looks fantastic!  I would be very happy to have this contribution.

EricWittmann:
If you want, we can create a separate i18n branch and merge it all in at once later, or you can continue updating this PR.   How do you want to proceed?

turutosiya:
@EricWittmann 
> If you want, we can create a separate i18n branch and merge it all in at once later
That sound good to me.

turutosiya:
@EricWittmann Please let me know when you have done creating i18n branch?

EricWittmann:
Whoops!  Sorry for the delay on this:

https://github.com/Apicurio/apicurio-studio/tree/feat/i18n

Feel free to open up PRs against that branch in whatever granularity suits you.

PR #1443, opened 2021-03-11
Fix command used in AsyncAPI deleteChannel: should be createDeleteChannelCommand

DyspC:
Hello, 

When deleting a channel from an AsyncAPI document, a DeletePathCommand is stacked in storage and causes errors on application (can't cast AaiDocument to OasDocument).

This PR changes the created command to DeleteChannelCommand.

--
The renameChannel uses a wrong command too but I think the modal is inactive and no fitting command exists in data-models yet.


EricWittmann:
Thanks!  FYI @lbroudoux

EricWittmann (approved):
tyvm

DyspC:
@lbroudoux if you hold a list of the missing asyncapi features and would like some help on it I may get some time from work so just let me know

lbroudoux:
Thanks for the fix. I should have too fast on testing that. I had the following remaining topics for AsyncAPI editor on top of my head:

- Servers support
- Bindings support (at least at the most frequent levels that are server, channel and operation)
- Traits management (message level first, operation level then)
- Payload type inline definition

I would be happy to have a quick chat to discuss that in depth (and we could even speak french 😉)

Cheers,

PR #1463, opened 2021-03-31
[i18n] Add Japanese compilation config

turutosiya:


turutosiya:
Followed by https://github.com/Apicurio/apicurio-studio/pull/1466

PR #1500, opened 2021-05-09
Azure SQL Support 

Surbhi27946:


back-end/hub-core/src/main/java/io/apicurio/hub/core/storage/jdbc/CommonSqlStatements.java
```diff
@@ -138,28 +142,46 @@ public String deleteLinkedAccounts() {
...
+            } else {
+                return "SELECT TOP 5 c.design_id, MAX(c.version) AS version "
+                        + "FROM api_content c "
+                        + "JOIN acl a ON a.design_id = c.design_id "
+                        + "WHERE a.user_id = ? AND c.created_by = ? "
+                        + "GROUP BY c.design_id "
+                        + "ORDER BY version DESC ";
+            }
+        }
+
```
EricWittmann, line 184: For any AzureSQL specific SQL statements, you should simply override the appropriate method in your new `AzureSQLStatements` class rather than have an `if` statement here.  Could you move all the appropriate azure-specific sql statement changes into that class?

EricWittmann (changes requested):
Thanks for this!  I have some questions.

1. Is this actually AzureSQL specific or will it work using any SQLServer instance?
2. Are we actually allowed to check into the SQL Server jdbc driver?  I'm concerned about licensing of that component.  I'm not sure it's a good idea even if the licensing is allowed.  I don't think we checked in the mysql or pg drivers, did we?

PR #1531, opened 2021-05-31
add java 11 in dockerfile because UnsupportedClassVersionError

vvilerio:
I had the following problem :

```
Caused by: java.lang.UnsupportedClassVersionError: Failed to link io/apicurio/studio/fe/servlet/servlets/AngularServlet 
(Module "deployment.apicurio-studio-ui.war" from Service Module Loader): 
io/apicurio/studio/fe/servlet/servlets/AngularServlet has been compiled by a more recent version of the 
Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

We can see that the release has been switched to Java 11 (see verify.yaml)
 
So in the thorntail images we need java 11
https://github.com/Apicurio/apicurio-studio/blob/master/.github/workflows/release.yaml#L110 
the image used in release.yaml is platforms/thorntail

### DIFF
--- a/platforms/thorntail/api/Dockerfile
+++ b/platforms/thorntail/api/Dockerfile
@@ -2,7 +2,7 @@ FROM centos:7
 LABEL authors="Eric Wittmann <eric.wittmann@redhat.com>"
 ENV RELEASE_PATH target/apicurio-studio-api-thorntail.jar
 RUN yum install -y \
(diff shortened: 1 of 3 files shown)


EricWittmann (approved):

EricWittmann:
Good catch thanks!  @carlesarnal can you double check the Quarkus images?

PR #1537, opened 2021-06-03
Revert mysql driver version to 22

carlesarnal:
See [this](https://bugs.mysql.com/bug.php?id=102435) for more information.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,7 @@
         <version.org.json>20210307</version.org.json>
         <version.org.jsonschema2pojo>1.1.1</version.org.jsonschema2pojo>
         <version.org.keycloak>11.0.3</version.org.keycloak>
-        <version.mysql>8.0.25</version.mysql>
+        <version.mysql>8.0.22</version.mysql>
         <version.org.postgresql>42.2.20</version.org.postgresql>
         <version.org.slf4j>1.7.30</version.org.slf4j>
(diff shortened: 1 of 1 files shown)


EricWittmann:
Looking at that defect you linked, should we keep the newer driver but change our code?  I assume the root of the problem is this:

https://github.com/Apicurio/apicurio-studio/blob/dfc277b03032562e4a3d37f16c851159340495b0/back-end/hub-core/src/main/java/io/apicurio/hub/core/storage/jdbc/mappers/ApiDesignChangeRowMapper.java#L49

We could change that, no?

carlesarnal:
> Looking at that defect you linked, should we keep the newer driver but change our code? I assume the root of the problem is this:
> 
> https://github.com/Apicurio/apicurio-studio/blob/dfc277b03032562e4a3d37f16c851159340495b0/back-end/hub-core/src/main/java/io/apicurio/hub/core/storage/jdbc/mappers/ApiDesignChangeRowMapper.java#L49
> 
> We could change that, no?

Yes, we need to adapt our code in custom `RowMapper` implementations. That said, this is still happening in some mappers inside jdbi, that's the reason for revert the connector version to the previous one. I have opened an [issue](https://github.com/jdbi/jdbi/issues/1880) in jdbi to keep track of this.

EricWittmann:
OK cool thanks for being on top of this.

PR #1557, opened 2021-06-21
blank kc real fixed

AlexBe1ov:
w/o realm var we will get blank kc realm and broken apicurio auth


carlesarnal:
hi @AlexBe1ov, have you tested this? afaik this was done this way since we have some defaults in place for this value. This distro has been tested and it's supposed to be working. Can you confirm that it's not working without the value?

carlesarnal:
This wouldn't hurt so I'm going to approve and merge. I don't think it's required, but it's probably better to have the realm set explicitly in the env template.

PR #1559, opened 2021-06-22
Fix missing // in default studio.keycloak.url

DyspC:
Hello, 

We're using the platforms/quarkus package to test locally (eclipse + wildfly + jboss plugins were a bit too much for us) and this lil miss caused us some errors since quarkus failed initializing the default tenant and only logs the error as debug :)

---
If you're interested in the alternative dev mode it uses some javaenv+sed shenanigans but I pushed it there https://github.com/Apicurio/apicurio-studio/compare/master...DyspC:dev-mode

### DIFF
--- a/platforms/quarkus/api/src/main/resources/application.properties
+++ b/platforms/quarkus/api/src/main/resources/application.properties
@@ -22,7 +22,7 @@
 quarkus.http.cors=true
 
 
-studio.keycloak.url=${APICURIO_KC_AUTH_URL:http:localhost:8090}
+studio.keycloak.url=${APICURIO_KC_AUTH_URL:http://localhost:8090}
 studio.keycloak.realm=${APICURIO_KC_REALM:apicurio}
 
 quarkus.oidc.application-type=service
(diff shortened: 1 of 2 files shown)


carlesarnal (approved):
Ah, yes, sorry, I tend to override this value with the environment variable. And please, give us any feedback you have with the Quarkus deployment! I will be more than happy to discuss any comments you might have.

carlesarnal:
Please, if that's ok, open a pr with the dev mode and I will look at it. Thanks again!.

DyspC:
Unfortunately we're not deploying the quarkus build yet since I had some issues I can't remember exactly and managed to bypass my CI's artifact size limit.

I'm taking note to contact you if I try again in the next few months

DyspC:
@carlesarnal Sorry I just noticed I forgot fixing the url in the websocket's properties

carlesarnal:
Hi @DyspC, for the websockets app you don't need to configure the auth server.

DyspC:
Oh yes sorry I'm not done waking up it seems

PR #1560, opened 2021-06-22
Alternative dev mode proposition using quarkus

DyspC:
Hello, 

This is a proposition of an alternative studio dev mode for those who are less prone to run their projects from their IDE or just don't want to use eclipse :)

This shouldn't be merged as is I think because all dev mode manuals should be at the same place and the example env file holds a configuration key from my registry source branch. I also wanted a direct and minimal link to send so I made it a separate file in our repo.

My launch commands are building -D options for maven instead of using the system env because the studio configuration handles both but some options (unless I missed something big) have to be set in the JVM

For example a redacted list of the options my team uses for the api:
```properties
# apicurio
apicurio.hub.storage.jdbc.init=true
apicurio.hub.storage.jdbc.type=postgresql9
apicurio.registry.api.url=http://my.apicurio-registry.example.com:8080/api
# jvm
jdk.http.auth.tunneling.disabledSchemes=
http.proxyHost=<corporate proxy setting>
http.proxyPort=<corporate proxy setting>
http.proxyUser=<corporate proxy setting>
http.proxyPassword=<corporate proxy setting>
https.proxyHost=<corporate proxy setting>
https.proxyPort=<corporate proxy setting>
https.proxyUser=<corporate proxy setting>
https.proxyPassword=<corporate proxy setting>
```
---
This is the dev mode I talked about in https://github.com/Apicurio/apicurio-studio/pull/1559


platforms/quarkus/ws/src/main/resources/application.properties
```diff
@@ -9,6 +9,7 @@
 %dev.quarkus.datasource.jdbc.min-size=2
 
 
+%dev.quarkus.http.port=${HTTP_PORT:8080}
```
EricWittmann, line 12: Should we default this one to another port?  The API is also defaulted to 8080.

EricWittmann:
Looks like a reasonable approach to me.  I agree we need easier ways to get people started with development that don't require Eclipse.  With the new Quarkus support, dev mode is even better because Quarkus will detect changes and restart, something Thorntail wouldn't do.  So for that reason IDE integration is less needed.  Although I still like launching the processes from Eclipse so I don't have to separately attach a debugger. :)

EricWittmann:
Thoughts @carlesarnal ?

DyspC:
The hot reload is nice to have but iirc there was an issue with a metric (api_calls or along those lines) that throws on reload in the api because its name is already registered.

platforms/quarkus/ws/src/main/resources/application.properties
carlesarnal: +1

platforms/quarkus/api/src/main/resources/application.properties
```diff
@@ -10,6 +10,7 @@
 %dev.quarkus.oidc.auth-server-url=${APICURIO_KC_AUTH_URL:http://localhost:8090/auth/realms/apicurio}
 %dev.quarkus.oidc.client-id=${APICURIO_KC_CLIENT_ID:apicurio-api}
 
+%dev.quarkus.http.port=${HTTP_PORT:8080}
```
carlesarnal, line 13: I don't see much value in adding this since this is already possible by using `QUARKUS_HTTP_PORT` or -Dquarkus.http.port.

platforms/quarkus/ws/src/main/resources/application.properties
carlesarnal: Same here, I don't think this is adding something new.

carlesarnal:
Right now I see two problems here. The first, and the most important one, is that the way this is starting the Quarkus applications the api and the ws will not share the same database and that's mandatory in Studio.

The second one, and is a minor thing, is the one that I have commented already, I don't think addding the ports to the `application.properties` adds much value since that's already possible using standard Quarkus env vars.

carlesarnal:
@EricWittmann we should probably add some tests that are based on the Quarkus distribution so the CI build fails if something breaks it. That's probably on me, I will create an issue and continue from there.

DyspC:
> Right now I see two problems here. The first, and the most important one, is that the way this is starting the Quarkus applications the API and the WS will not share the same database and that's mandatory in Studio.

In this case shouldn't the default DB settings in both application.properties use an external db like this?

```properties
%dev.quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/apicuriostudio
%dev.quarkus.datasource.db-kind=postgresql
%dev.quarkus.datasource.username=apicurio
%dev.quarkus.datasource.password=apicurio
```

along with those options in the default .javaenv files
```properties
apicurio.hub.storage.jdbc.init=true
apicurio.hub.storage.jdbc.type=postgresql9
```

> The second one, and is a minor thing, is the one that I have commented on already, I don't think adding the ports to the `application.properties` adds much value since that's already possible using standard Quarkus env vars.

I didn't know about those options, I'll change the recommendation to use QUARKUS_HTTP_PORT on the command line instead

carlesarnal:
> > Right now I see two problems here. The first, and the most important one, is that the way this is starting the Quarkus applications the API and the WS will not share the same database and that's mandatory in Studio.
> 
> In this case shouldn't the default DB settings in both application.properties use a shared db like this?
> 
> ```ini
> %dev.quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/apicuriostudio
> %dev.quarkus.datasource.db-kind=postgresql
> %dev.quarkus.datasource.username=apicurio
> %dev.quarkus.datasource.password=apicurio
> ```
> 
> along with those options in the default .javaenv files

Yes, exactly, you will need to override that config, but I do no see that change in this dev mode, that's why I'm saying it's a problem. You can modify the URL by using QUARKUS_DATASOURCE_JDBC_URL in both API and WS.

> 
> ```ini
> apicurio.hub.storage.jdbc.init=true
> apicurio.hub.storage.jdbc.type=postgresql9
> ```
> 
> > The second one, and is a minor thing, is the one that I have commented on already, I don't think adding the ports to the `application.properties` adds much value since that's already possible using standard Quarkus env vars.
> 
> I didn't know about those options, I'll change the recommendation to use QUARKUS_HTTP_PORT on the command line instead

Yes, as a note, every property in quarkus is exposed as an env var. In this case, as you can see, `quarkus.http.port property` becomes `QUARKUS_HTTP_PORT`.

EricWittmann:
Good point about the shared database!

We can have default H2 settings for dev mode that will result in a shared DB.  I have an example of this somewhere, let me dig it up and report back.

EricWittmann:
OK so we can set the value of `%dev.quarkus.datasource.jdbc.url` to `jdbc:h2:~/data/studio;AUTO_SERVER=TRUE` for both services and they will share a DB.  Whichever one starts up first will create the DB (in-memory) and also open a port for the second service to connect to when it starts.

EricWittmann:
Looking good.

carlesarnal:
With the latest changes to the db, I think we're getting close to a good dev mode. @DyspC anything else you want to address? Please, when this is fully ready, mark it as ready for review and I will try it locally.

Thanks for your contribution!

EricWittmann:
@carlesarnal This has been sitting for too long - what do you think?  Ready to review/merge?

carlesarnal:
> @carlesarnal This has been sitting for too long - what do you think? Ready to review/merge?

Yes, I think this is good to merge. I was just waiting confirmation from @DyspC.

DyspC:
Sorry I forgot about this one and just got off for a week

- This dev mode depends on the jars for API and WS having been built, I don't know if there's a way to change this but a maven clean install excluding the front end and platforms runs in about 90s here
- My launch command had an issue where in some cases the last line of javaenvs was skipped I'll commit the fix when I'm back
- Perhaps the `.javaenv` files being commited and editable means they should have a non hidden name (`java.env` or a plain `javaenv`)

DyspC:
> * This dev mode depends on the jars for API and WS having been built, I don't know if there's a way to change this but a maven clean install excluding the front end and platforms runs in about 90s here
> * My launch command had an issue where in some cases the last line of javaenvs was skipped I'll commit the fix when I'm back
> * Perhaps the `.javaenv` files being commited and editable means they should have a non hidden name (`java.env` or a plain `javaenv`)

I updated those 3 points (and discovered the maven -am flag that ought to save me some time 😅)

EricWittmann (approved):

EricWittmann:
Thanks!

PR #1578, opened 2021-07-02
Upgrade Angular to v11, Node.js to v14

zregvart:
In https://github.com/syndesisio/syndesis we embed Apicurio Studio and we would like to upgrade Node.js from v10 to latest LTS, which is v14 currently. When upgrading to Node.js v14 we have found an issue with an older version of node-sass brought about as a transitive dependency of Apicurio Studio.
This made me look into upgrading Apicurio Studio to newer versions of Node.js, and consequentially Angular. This was not as difficult to do and I managed to upgrade to version 11, stable version currently being v12. Until https://github.com/maxisam/ngx-clipboard and https://github.com/valor-software/ngx-bootstrap support Angular v12, v11 is the newest version that Apicurio Studio can be upgraded to.

I've managed to get the project to compile, and I'm going to do some testing to see if there are any issues; mostly focusing on the embedded use in Syndesis.

I do realize that this is a big set of changes, I've tried to break down the work in many commits, each corresponding to the one migration step done [Angular Update Guide](https://update.angular.io/?v=7.2-11.0) with addition to any manual migrations/fixes. Hopefully that would make the review a bit easier.

Happy to adjust this or fix any issues brought on by this. I see that I need to rebase, and I still would like to perform a bit of manual testing so marking this as a draft.


zregvart:
This looks good now, last issue I found and fixed in 1fa007b was a blocker. I'm having some issues with embedding as something changed with CSS. I think this needs to be fixed on the Syndesis side, it's about how CSS file is included, what we had no longer works. The UI shows up, albeit with styling issues, and seems functional.

zregvart:
There is one more issue similar to the one I fixed in 1fa007b, I'll try to get that fixed soon.

zregvart:
> There is one more issue similar to the one I fixed in [1fa007b](https://github.com/Apicurio/apicurio-studio/commit/1fa007bf5ef85308d3179898f5f8cb05caefb5cd), I'll try to get that fixed soon.

Turns out this was the same issue as in 1fa007b but needed to be fixed in Syndesis as well.

I think this is ready for the review.

EricWittmann:
Great work, thanks very much @zregvart !  I have done some testing locally in both dev mode and prod mode and everything seems to work just fine. :)

PR #1588, opened 2021-07-12
Adjust Postgres Image for Openshift Template Closes #1505

### LINKED ISSUE #1505: Standalone Template failing to deploy Postgres
Hi,

I was just deploying Apicurio Studio on Openshift 4.6 Cluster using the standalone template and the postgresql pod never started because of unresolved image, I think that it's because of this spec on the pod template:
```yaml
 containers:
          - name: postgresql
            image: centos/postgresql-95-centos7
            ports:
              - containerPort: 5432
                protocol: TCP
            env:
```
I got it working by changing it to this:
```yaml
containers:
          - name: postgresql
            image: postgresql:9.6-el8
            ports:
              - containerPort: 5432
                protocol: TCP
            env:
```

Is this worth a PR? I think that the image could be a template Parameter.

I can make this modifications if you guys think this should be this way.

Regards

raffamendes:
Changed the Postgresql Image to be a template parameter as it was not deploying correctly with the old values in Openshift 4.7.

Also added a default value for the image tag that resolve to a existent tag that exists in the OpenShift namespace.

it also closes issue #1505

### DIFF
--- a/distro/openshift/apicurio-standalone-template.yml
+++ b/distro/openshift/apicurio-standalone-template.yml
@@ -242,7 +242,7 @@ objects:
           from:
             kind: ImageStreamTag
             namespace: openshift
-            name: 'postgresql:9.5'
+            name: ${POSTGRESQL_IMAGE}
       - type: ConfigChange
     replicas: 1
     test: false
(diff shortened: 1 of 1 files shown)


carlesarnal (approved):

carlesarnal:
Thanks @raffamendes

Issue #1562, opened 2021-06-23
Configurable API templates in Apicurio Studio


DyspC:
Hello, 

We're deploying apicurio to edit both OAS & AAI APIs and we'd like to give our users custom templates with for example our kafka brokers or imports of our message traits / shared error payloads.

Rebuilding docker images from source to allow custom templates does not feel right so we'll work on a way to load them later instead of hardcoding them.

We've discussed a bit those two ideas are sticking:
- Extending the docker image to copy a template folder with all metadata and API content
- Adding a configuration key like apicurio-ui.templates.url that points to an array like [this one](https://github.com/lbroudoux/apicurio-studio/blob/fb8d0b6974258c2e1359ee28fdce98b7653356da/front-end/studio/src/app/services/template.service.ts#L36)

Both solutions will probably need an additional flag to state whether the custom templates should replace the default ones or extend the selection and serving the templates through one of the backends would allow using some cache.

I don't think we are the only ones with this specific need so we should discuss the best way to handle this and make it worth a PR.

I have a personal preference for the configuration keys because it does not force a parallel docker release train

EricWittmann:
I like this idea very much.  I think we should completely remove the hard-coded array of templates from the UI layer and instead serve the list of them from the back-end.

I don't have a strong opinion between the template folder and the conf key (probably a preference for the config key for the reason already given).  Another option, however, would be to update the back-end to manage the collection of templates as a resource in the REST API.  This would allow dynamic configuration of these templates and management of them via an API (and eventually a UI).

In any case, I think the first step would be to at least expose a REST API endpoint that would allow the UI to query for the list of templates.  That API can (at first) be read-only and driven from config.  And we can then enhance it in the future to allow full dynamic management.

DyspC:
The REST API approach is nice indeed. Although this calls for user roles and scoping (user, team, instance...) since we don't want anybody to create instance wide templates.

With a new `Template` ApiContentType that holds a reference to a `Document` row of the same design_id (only the version would be needed then) this could fit in the existing DDL.

This would work for user defined templates but will need a bit of rework when (if?) scopes are introduced (I think the json stored in a character varying column won't allow postgresql's JSONB operations and I don't know what mysql and h2 have to offer to avoid application side filtering)

I don't know how it can be translated for default templates though (forcing them into the DB with a fake user is hack-ish and will cause problems with upstream updates)

EricWittmann:
[...]

PR #1595, opened 2021-07-16
Move API template delivery from studio-fe-studio to studio-be-hub-api

DyspC:
Hello, 

This is a first part for #1562, moving the static templates from the frontend to the backend.


front-end/studio/src/app/services/template.service.ts
```diff
@@ -17,1662 +17,55 @@
...
+     */
+    constructor(http: HttpClient, authService: IAuthenticationService, config: ConfigService) {
+        super(http, authService, config);
     }
 
     /**
      * Gets a list of templates for the given spec version.
      * @param type
      */
     public getTemplates(type: string): ApiDesignTemplate[] {
```
EricWittmann, line 49: I think this needs to return a `Promise<ApiDesignTemplate[]>` if we're going to possibly be fetching the data from the server now (when not cached).

back-end/hub-api/src/main/java/io/apicurio/hub/api/rest/impl/TemplatesResource.java
```diff
@@ -0,0 +1,1655 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * @author c.desc2@gmail.com
+ */
+public class TemplatesResource implements ITemplatesResource {
+
+    private static Logger logger = LoggerFactory.getLogger(TemplatesResource.class);
+
+    private static final List<ApiTemplate> allTemplates = List.of(
```
EricWittmann, line 35: Let's store each template as a resource instead of inlining them in the Java code.  We can load the resources in a `static` intializer.

EricWittmann (changes requested):
Looks good!  I had a couple of comments, let me know what you think.

DyspC:
I'm not sure about the way to handle promises but wdyt?

EricWittmann:
Awesome @DyspC !  This is great, almost perfect.  The only change I would ask for is to keep the `templateCache` as a cache of "" instead of caching the `Promise`.  Instead, the logic of the method should check the cache, and use `Promise.resolve()` in the case of a cache hit.  Something like this:

```typescript

    private templateCache: { [key: string]: ApiDesignTemplate[] } = {};

    public getTemplates(type: string): Promise<ApiDesignTemplate[]> {
        if (!this.templateCache.hasOwnProperty(type)) {
            console.info(`[TemplateService] Templates of type '${type}' are not ready, loading them from the API...`)
            let options: any = this.options({ "Accept": "application/json" });
            // Fetch the template from the server, return a Promise since the REST call is async.
            return this.httpGet<any>(this.endpoint("/templates", {}, { type: type }), options)
                .then(rawTemplates => {
                    let rval: ApiDesignTemplate[] = [];
                    for (let template of rawTemplates) {
                        rval.push({
                            type: template.type,
                            name: template.name,
                            description: template.description,
                            content: template.document
                        })
                    }
                    
                    // Cache the templates for this type.
                    this.templateCache[type] = rval;
                    return rval;
                });
        }
        return this.templateCache[type];
    }

```

DyspC:
I'm not sure, I copy pasted this but it makes the tab poll for templates until it crashes (cache hits but nothing is displayed).

EricWittmann:
@carlesarnal if you have some time in the next 2 weeks can you help with this?  @DyspC I'm going to be on vacation for a bit, hopefully Carles can help out!  :)

EricWittmann:
FYI I'm back from PTO and trying to get some eyeballs on this, sorry for the delay @DyspC

carlesarnal:
Hi, @DyspC I have tested this locally in both chrome and firefox and it's working fine for me. Do you have more info about your setup? I'm testing using the Quarkus build.

carlesarnal:
> Hi, @DyspC I have tested this locally in both chrome and firefox and it's working fine for me. Do you have more info about your setup? I'm testing using the Quarkus build.

Just FYI, I have also tested this using the quickstart and it also works.

carlesarnal:
Hi, as a recap,

I've been able to reproduce the issue, I wasn't able to do it due to a code mismatch.

I have solved the problem by propagating the array of elements instead of the promise itself. I think this is the right thing to do. If we cache elements, we propagate elements, if we cache promises, then we propagate promises.

DyspC:
> > Hi, @DyspC I have tested this locally in both chrome and firefox and it's working fine for me. Do you have more info about your setup? I'm testing using the Quarkus build.
> 
> Just FYI, I have also tested this using the quickstart and it also works.

Were you testing the raw branch or after applying Eric's suggestion? I just tried the patch again with Brave/Safari and my dev mode / the quickstart and both² get the tab hang issue.

Also it looks like 5ad41d6 and b585e61 cancel themselves so we're back to this remark https://github.com/Apicurio/apicurio-studio/pull/1595#discussion_r673342029

carlesarnal:
> > > Hi, @DyspC I have tested this locally in both chrome and firefox and it's working fine for me. Do you have more info about your setup? I'm testing using the Quarkus build.
> > 
> > 
> > Just FYI, I have also tested this using the quickstart and it also works.
> 
> Were you testing the raw branch or after applying Eric's suggestion? I just tried the patch again with Brave/Safari and my dev mode / the quickstart and both² get the tab hang issue.
> 
> Also it looks like [5ad41d6](https://github.com/Apicurio/apicurio-studio/commit/5ad41d66edb42ce3902dc0b13239e7e19c7c934b) and [b585e61](https://github.com/Apicurio/apicurio-studio/commit/b585e61fdba4be8bf6b9a87f0fb109a0bc93155b) cancel themselves so we're back to this remark [#1595 (comment)](https://github.com/Apicurio/apicurio-studio/pull/1595#discussion_r673342029)

I was testing the raw branch, that's why I wasn't getting the error, after applying the patch I'm getting the error.

EricWittmann:
OK everyone - sorry about this.  It has been a little while since I've done any Angular development, and I forgot how it works.  :)  I've made some changes so that the templates are loaded asynchronously during the "LoadAsyncPageData" phase of the page load.  So now the form component just gets the templates from the template service and doesn't deal with promises or anything.  The template service caches the templates in a "load" step, which is triggered (and waited on) by the Create page's async load phase.

My suggestion did not work due to how angular renders a page - specifically how it re-renders a page!  As a result, angular was just calling the `templates()` method again and again.  Because each time it called that method, a new `Promise` was being returned (my bad on that).  Caching the `Promise` instead of the data was actually working.

In any case, the approach I"m going with is, I think, better in the specific context of the Studio application (not necessarily Angular in general).  :)

EricWittmann:
Thank you @DyspC for the contribution - sorry it's taken so long to get it merged!

EricWittmann:
If interested, here is my change:  https://github.com/Apicurio/apicurio-studio/commit/38a8609513cb173ad357292bcec9f59f38365b7b

EricWittmann:
It includes a UI spinner in case the loading of the templates takes some time for some reason.

PR #1601, opened 2021-07-20
Use entrypoints instead of cmds in thorntail images

DyspC:
Hello, 

I'm adding a corporate secret management layer to our studio deployment and I came into this issue where extending the image with an entrypoint erases the parent CMD

This change should allow image extension without having to maintain a copypaste of the CMD layer (eg. with another entrypoint ending with `exec /opt/apicurio/entrypoint.sh $@`)

### DIFF
--- a/platforms/thorntail/api/Dockerfile
+++ b/platforms/thorntail/api/Dockerfile
@@ -8,10 +8,12 @@ RUN yum install -y \
 WORKDIR /opt/apicurio
 
 ADD ${RELEASE_PATH} /opt/apicurio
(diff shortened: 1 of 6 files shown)


EricWittmann:
LGTM wdyt @carlesarnal ?

carlesarnal:
LGTM, the only question for me is if any of you have tested this locally or if I need to test them since workflows don't cover docker images.

DyspC:
I'm running an image based on this branch on a staging env (with my added layer / new entrypoint) and no issue whatsoever

carlesarnal (approved):

PR #1641, opened 2021-08-27
Fix Channel source edition / clone and name regexp

### LINKED ISSUE #1626: AsyncAPI: Editing a channel in source editor causes a circular reference
Hi,

We got trouble when editing an AsyncAPI channel from the source view. The save button triggered a circular reference issue, as below: 

image

Thanks,

DyspC:
Closes #1626

This fixes the command used to save changes in the channel source editor

I also added an implementation for cloning channels and extended the regex for channel names (it wasn't possible to create a channel named user/signedup)

https://www.asyncapi.com/docs/specifications/v2.0.0#channelsObject

Please note I could only test the regexes in the thorntail build / dev mode / sourcing the .env files so there could be escape issues with the other distros, I listed test cases [there](https://regexr.com/64hp3)


EricWittmann:
LGTM

PR #1679, opened 2021-09-27
Introduce user role injection from keycloak

DyspC:
According to https://github.com/Apicurio/apicurio-studio/issues/1562#issuecomment-926855759, here is the first part that supports keycloak adding roles to the token and makes them accessible from the studio components.

I also added an AuthorizationService class to host auth decision depending on the user's ACLs / roles.

The role definition to update the realm of the apicurio-studio-auth image 
```json
{
  "id": "79f49397-7fb2-40b8-9e4c-bac26d756ae7",
  "name": "apicurio_admin",
  "description": "Admin role for users in Apicurio Studio",
  "composite": false,
  "clientRole": false,
  "containerId": "apicurio"
}
```

To get roles in the front end in dev mode, add them directly to the config.js file 
```
    user : {
        "login" : "[redacted]",
        "id" : 0,
        "name" : "[redacted]",
        "email" : "[redacted]",
        "roles" : [ "APICURIO_ADMIN" ]
    },
```

This is was originally part of a wider exploration to create an admin role with the ability to use an instance as if it had share with everyone enabled (in our use case a team that helps other people in the company design their APIs and manages API tooling) but I'm not satisfied with how hack-ish it is 
[Raw apicurio god mode role.txt](https://github.com/Apicurio/apicurio-studio/files/7236687/Raw.apicurio.god.mode.role.txt)

### DIFF
--- a/back-end/hub-api/pom.xml
+++ b/back-end/hub-api/pom.xml
@@ -137,6 +137,11 @@
             </exclusions>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>javax.json</groupId>
+            <artifactId>javax.json-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
 
         <!-- Keycloak -->
         <dependency>
--- a/back-end/hub-api/src/main/java/io/apicurio/hub/api/rest/impl/DesignsResource.java
+++ b/back-end/hub-api/src/main/java/io/apicurio/hub/api/rest/impl/DesignsResource.java
@@ -88,6 +88,7 @@ import io.apicurio.hub.api.microcks.IMicrocksConnector;
 import io.apicurio.hub.api.microcks.MicrocksConnectorException;
 import io.apicurio.hub.api.rest.IDesignsResource;
 import io.apicurio.hub.api.security.ISecurityContext;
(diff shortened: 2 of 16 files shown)


DyspC:
I don't know where client access roles are stored in the token and honestly don't know what they are so I won't have any opinion against that

carlesarnal:
> I don't know where client access roles are stored in the token and honestly don't know what they are so I won't have any opinion against that

Roles in keycloak has a level, so you have realm roles and client roles. It's a matter of configuration to put the client roles in the token or not. But never mind, we can add that later if we think it's worth it.

Thank you very much for the contribution.

DyspC:
> OK that's great, thanks. Perhaps a feature flag for this, with a default value of `false`? And then we can separately address paging in the list views (which is something that should be done anyway).

I'm digging this comment out because our API admins started requesting a way to see all specifications :)

Did you say knowing what weird hacks are in the txt patch of the original comment?

How about using the roles for this instead of the whole instance configuration? Perhaps putting the template edition under the `apicurio_admin` instead of a named role wasn't the best choice, it should probably be composite to enable all role base features and keep individual role features assignable one by one

PR #1692, opened 2021-10-14
Add a dereference option to the git publish flow

### LINKED ISSUE #1418: Feature Request: Allow dereferencing imported types when publishing to git
The import feature for components is great, because it allows maintaining well structured API-definitions while at the same time preventing repetition.
Unfortunately the resulting definition files are invalid Open API defitions, because the cross-api references are unresolvable outside of apicurio. (Compare #1095)
To mitigate this, there is an option to dereference all imported components when downloading an api definition file.

This option should also be available when publishing an api definition to git. That way the import feature can be used together with a git-based workflow.

DyspC:
Fixes #1418 

Hello, 

This PR adds a drop down menu to the publish page to select whether or not references should be resolved (The same as in the download modal). 
I also refactored a bit the `publishApi` method to use `getApiContent` because both did the same manipulations on the document.

image


EricWittmann:
Nice change, thanks @DyspC .  I've merged this using the CLI so I could make some minor tweaks.  Not sure why GH didn't figure that out.  Closing as merged.

PR #1728, opened 2021-11-25
Follow up for #1679, adds null safety for realms with no default roles

### LINKED ISSUE #1724: NullPointerException when user has no keycloak roles
In the last release of apicurio, 0.2.51, a mapping from keycloak realm roles was added at https://github.com/Apicurio/apicurio-studio/blob/master/front-end/servlet/src/main/java/io/apicurio/studio/fe/servlet/filters/KeycloakAuthenticationFilter.java#L81

For us the users don’t have any realm roles and this seems to lead to `token.realmAccess().getRoles()` being `null`. After giving the relevant users any random role, apicurio works again.

Keycloak version 12.

DyspC:
fixes #1724


carlesarnal (approved):

carlesarnal:
Thanks for spotting this and fixing it.

PR #1736, opened 2021-12-06
Fix templates and configs for Kubernetes and Helm

craicoverflow:
I experienced compatibility issues when trying to deploy Apicurio to Kubernetes using Minikube so I decided to fix them. 

Perhaps you require these to be on an older version and need them to stay as they are, in which case feel free to close this.

### DIFF
--- a/distro/kubernetes/apicurio-studio-ui-deployment.yaml
+++ b/distro/kubernetes/apicurio-studio-ui-deployment.yaml
@@ -67,7 +67,7 @@ spec:
         - name: JAVA_TOOL_OPTIONS
           value: -Djava.net.preferIPv4Stack=true
         - name: APICURIO_UI_VALIDATION_CHANNELNAME_REGEXP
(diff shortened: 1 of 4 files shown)


distro/kubernetes/apicurio-studio-ui-deployment.yaml
```diff
@@ -67,7 +67,7 @@ spec:
         - name: JAVA_TOOL_OPTIONS
           value: -Djava.net.preferIPv4Stack=true
         - name: APICURIO_UI_VALIDATION_CHANNELNAME_REGEXP
-            value: '{([^\x00-\x20\x7f"\'%<>\\^`{|}]|%[0-9A-Fa-f]{2}|\{[+#./;?&=,!@|]?((\w|%[0-9A-Fa-f]{2})(\.?(\w|%[0-9A-Fa-f]{2}))*(:[1-9]\d{0,3}|\*)?)(,((\w|%[0-9A-Fa-f]{2})(\.?(\w|%[0-9A-Fa-f]{2}))*(:[1-9]\d{0,3}|\*)?))*\})*'
```
craicoverflow, line 70: I got `error: error parsing apicurio-studio-ui-deployment.yaml: error converting YAML to JSON: yaml: line 71: found character that cannot start any token` because `'` was not escaped correctly. So I _think_ I have keept the regex pattern the same, but maybe you could check.

EricWittmann (approved):

EricWittmann:
Thanks!

PR #1765, opened 2021-12-23
fix

fari-99:
- delete line `apicurio-studio-db:` on docker-compose.apicurio.yml (error docker compose not valid)
- fix generate random number so can be used on mac ([source](https://unix.stackexchange.com/questions/45404/why-cant-tr-read-from-dev-urandom-on-osx))
- fix where env on setup for mysql root password not the same as on `docker-compose-as-mysql.yml` (`AS_MYSQL_ROOT_PASSWORD`)


EricWittmann:
Thanks!

PR #1796, opened 2022-01-25
Adjustments to deploy Quarkus flavor with Operator

lbroudoux:
Some adjustments to allow deployment of Quarkus flavor using Operator:
- Fix exposition port of ws component
- Force OIDC redirect_uri to be https when deployed behing an ingress controller that handles TLS
- Make Keycloak URL config the same that Wildfly falvor so that both can be deployed with Operator


carlesarnal (approved):
LGTM, thanks!

PR #1819, opened 2022-02-18
Helm chart improvements

mahgo:
Added improvements to the helm chart:
- Fixed a typo in the notes
- Added being able to use external db instead of installing one with chart
- Added ingress annotation and tls configuration
- Added deployment resource configuration
- Added being able to use an existing secret instead of creating a new one


EricWittmann (approved):

EricWittmann:
Thanks!

PR #1829, opened 2022-03-03
Add default text editor for unsupported/unknown artifact types

craicoverflow:
I've kept this in a draft for now, because it partially includes experimental changes which would reduce boilerplate, but also could be incompatible with our needs that I don't know about yet.

I've separate this PR into two distinct commits:

* 6f8696d99123198409843af0e8dd2d5f8055325b adds the basic text editor without much other changes
* 3482167f032f85cbed2d9b2fd20e6631029d628e adds a base editor component. The biggest change here which is an experiment is to see whether we can use functional components more as class-based components tend to be larger and have more boilerplate. I actually think we could eliminate the use of the custom `PureComponent` or at least limit its usage, by using custom hooks instead - this is why I have included the useNavigation (`ui-rhosr/src/app/hooks/useNavigation.ts`) as an example of how this could work. There could be other reasons I am not aware of yet that means the class-based components are needed, in which case that's totally fine, I just wanted to try this out.

So I can either:

   * Remove the second commit and let's keep it with classes for now
   * Merge it like this because the Editor component will not need navigation anyway (yet)


EricWittmann:
Regarding class vs. functional components - there is no requirement for class style components, I just like them better and find them easier to read and maintain.  But that's mostly just personal preference and comfort.  I'm **told** that functional components are superior, but I don't recall the arguments about why.  I don't have an ideological reason to use classes, it's just what I know and am used to.  I have seen some code, especially in app services, that uses functional components and I find harder to read/understand.  But again, possibly that's just due to lack of familiarity.

EricWittmann:
Otherwise LGTM :)

craicoverflow:
Thanks for the context. I do think this is down to familiarity more than anything. Some of the benefits of functional components are:

* Much smaller with less boilerplate
* The compiled code ends up being smaller
* State-management is more straightforward with hooks (again, these look super weird at first but become quite intuitive and flexible).

I don't want to impose a style that the team might not be comfortable with though, my own preference is functional but I've done a lot of React so am comfortable with both.

@carlesarnal - during your day of learning with React did you come across class versus functional components, or do you have your own preference?

carlesarnal:
> Thanks for the context. I do think this is down to familiarity more than anything. Some of the benefits of functional components are:
> 
>     * Much smaller with less boilerplate
> 
>     * The compiled code ends up being smaller
> 
>     * State-management is more straightforward with hooks (again, these look super weird at first but become quite intuitive and flexible).
> 
> 
> I don't want to impose a style that the team might not be comfortable with though, my own preference is functional but I've done a lot of React so am comfortable with both.
> 
> @carlesarnal - during your day of learning with React did you come across class versus functional components, or do you have your own preference?

Yes, I did some React. I don't have a strong opinion. I see why some prefer class-based, especially those with a java background, but most of the materials I found were using functional components.

PR #1873, opened 2022-04-07
fix: use keycloak-js from npm

craicoverflow:
Currently Studio expects you to drop in a keycloak.js file in order to use the UI. This likely was necessary at some point but not any more since keycloak-js is available as an npm dependency. We should use this dependency as it will ensure we stay up to date with the latest fixes and also simplifies getting set up.

The next step is to remove this step from the docs [here](https://apicurio-studio.readme.io/docs/setting-up-a-development-environment#download-keycloakjs). Where can I find the source?


EricWittmann (dismissed):

EricWittmann:
@craicoverflow build failure :(

craicoverflow:
> @craicoverflow build failure :(

Yeah, I requested review prematurely.

> Distributing npm packages with 'dependencies' is not recommended. Please consider adding keycloak-js to 'peerDependencies' or remove it from 'dependencies'.

Hadn't heard that before, interesting.

Issue #2082, opened 2023-01-09
too large websocket message cause offline


Handgrip:
When I manually edit the json document and save it, when the json document is too large, it will lead to offline(websocket no response).
![image](https://user-images.githubusercontent.com/30646834/211273759-74c7c7f0-f203-4818-92b4-8770234b6d65.png)

![image](https://user-images.githubusercontent.com/30646834/211273775-d204ed13-8517-4869-856a-c024908c02d3.png)

![image](https://user-images.githubusercontent.com/30646834/211273784-73ee19a3-fa01-477e-a519-40b29f7d2f1f.png)

Handgrip:
![image](https://user-images.githubusercontent.com/30646834/211274816-2173f50f-512b-4206-aa35-d4a1050aacf6.png)
 seems related to https://github.com/Apicurio/apicurio-studio/issues/2024, I'm using https://studio.apicur.io/, did it update to the last version?

blackjack4494:
Having the exact same issue here also on the official web version.  
When I try to add an example response it will immediately lose connection and won't reconnect no matter what I do. Even deleting the input won't let me reconnect.  
This already happens with smallish examples of around 1mb.

EricWittmann:
I thought this was fixed on https://studio.apicur.io/ in this change:  https://github.com/Apicurio/apicurio-studio/blob/master/platforms/quarkus/ws/src/main/resources/application.properties#L25

I believe the latest version with that change is deployed.  But apparently not.  I will investigate.

Note: another improvement that we should consider is to send a JSON Diff as the payload for a source code change, rather than the entire source.  Not sure if I'll ever reasonably have time to implement that though.

EricWittmann:
It seems that @romilin fixed this in https://github.com/Apicurio/apicurio-studio/pull/2124 - I apparently typed the property incorrectly in `application.properties`.  :(

I'll get a release done and deployed asap.

EricWittmann:
This should be finally fixed and released/pushed to studio.apicur.io.  I'm going to close this but please re-open if it is **still** not working!

PR #2124, opened 2023-03-07
Fix too large message cause offline

romilin:
Hello, 

In answer to issue #2082 : 
The correct spelling of the quarkus property websocket max frame size is : quarkus.websocket.max-frame-size

Official doc : 
image

Can you check this modification ? 
Happy if it helps :smile:


EricWittmann:
OMG good catch.  tyvm

Issue #2130, opened 2023-03-11
CORS policy issue after Quarkus update


v-bilous:
Hi @EricWittmann , how r u?

I see an issue related to the new version of Quarkus.
It's probably the new behavior for CORS settings.
When we have 
`quarkus.http.cors=true` only set
I see the same issue when I open UI  https://github.com/quarkusio/quarkus/issues/31523
<img width="1579" alt="Screenshot 2023-03-11 at 19 48 42" src="https://user-images.githubusercontent.com/19833841/224503883-fe777286-a09c-459c-bed3-ec598792512d.png">

`quarkus.http.cors.origins=*` is fix the issue for me.

Thanks,
Vova

adel-chouchane:
Hello @v-bilous , I am using the latest images of apicurio and i am having the same error, what is the env variable i need to pass to the UI ?

v-bilous:
hi @adel-chouchane , thank you! you are right, what I see it should be fixed in `apicurio-studio-ui`, but CORS rules were added to `apicurio-studio-api` only.
The change https://github.com/Apicurio/apicurio-studio/commit/9d94555a7a364e973e4a99fae6219d05cc246d80

@adel-chouchane you can take this image which is based on the latest version as a temporal solution `djetcloud/apicurio-studio-ui:1.0.3`, or just use the previous version `apicurio/apicurio-studio-ui:0.2.59.Final`

EricWittmann:
Hey everyone.  I don't think the UI project should need the CORS changes.  CORS enablement **should** only be needed on the REST API, not on the UI project.

I think my changes have fixed studio.apicur.io, and everything works locally for me now.  Please correct me if I"m wrong!

There is a new release 0.2.61.Final that you can try and let me know what happens.

EricWittmann:
This has been confirmed - it was working ok in Firefox but failing in Chrome and Safari.  I've merged the PR and I have a fix for the issue (env var config) ready for deployment in studio.apicur.io.  Once the ops team has a chance to merge that PR we should be good.  I'll close this as fixed and we can re-open it if it's still failing.

EricWittmann:
Thanks @v-bilous for the PR.

v-bilous:
Thank you for resolving it @EricWittmann

EricWittmann:
The 0.2.59 version still has the issue - can you try 0.2.61 instead?  If that doesn't work can you provide a little more information about exactly what is happening?

[...]

PR #2132, opened 2023-03-14
CORS config for UI

romilin:
Hello @EricWittmann ,

The recent modification about CORS seems to be missing something. 
I have just copy what you write in the API config and add it to the UI config. 

It solves this problem for me : 
image

### DIFF
--- a/platforms/quarkus/ui/src/main/resources/application.properties
+++ b/platforms/quarkus/ui/src/main/resources/application.properties
@@ -19,7 +19,12 @@
 %prod.quarkus.oidc.authentication.force-redirect-https-scheme=${APICURIO_OIDC_REDIRECT_HTTPS:true}
 %prod.quarkus.log.level=INFO
 
+# http
 quarkus.http.cors=true
+quarkus.http.cors.origins=${CORS_ALLOWED_ORIGINS:*}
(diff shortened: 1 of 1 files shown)


v-bilous (approved):

v-bilous (approved):

v-bilous:
Hello, this PR will fix the issue https://github.com/Apicurio/apicurio-studio/issues/2130

crmccar:
I believe this issue is completing blocking use of https://studio.apicur.io/. I could be wrong, but I'm stuck on the loading image and the CORS errors are the only thing showing up in the Chrome web development console.

EricWittmann:
See my comment on https://github.com/Apicurio/apicurio-studio/issues/2130 - I don't **think** this is needed.  But please correct me if I'm wrong and provide a configuration that's failing for you.  :)

EricWittmann:
Although I see from your screenshot that maybe you're getting CORS failures trying to load certain resources.

andapony:
I'm getting the same problem, both with Chrome and Safari.

EricWittmann:
You're right - I was using Firefox.  Chrome fails for me.

I'll merge this PR and fix the issue on studio.apicur.io (which doesn't require a new release).

EricWittmann (approved):

Issue #1887, opened 2022-04-25
Properties Section: Example 


lerio29:
the generated examples do not respect the rules  like minLength/maxLength

![image](https://user-images.githubusercontent.com/8198968/165116818-a03f3b70-5f82-4de4-bd9f-17ffe8376844.png)

EricWittmann:
Good point, thanks.  I've added it to the editor backlog.

pedrogclp:
I'll be watching this, it's a good point.

EricWittmann:
Fixed in https://github.com/Apicurio/apicurio-studio/pull/2140

PR #2140, opened 2023-03-20
Add example generator

romilin:
Hello, 

I saw the issue #1887. 
The issue is about using the defined properties rules in the exemple generated. 
I tried to add something to ExampleGenerator for that. 

What is different : the password, string, boolean and integer are generated randomly using the defined rules.

What do you think about this ?


EricWittmann (approved):
Looks great thanks for the contribution!

PR #2255, opened 2023-08-24
Example : fix bug string without maxLenght + add AllOf

romilin:
Hello, 

Here is two modifications on example generation : 

- Fix example bug when the string properties have no "maxLenght" attribute  -> Random length 
- When there is an "AllOf" inheritance, inherited properties are added to the generated example


EricWittmann:
Thanks!

Issue #2258, opened 2023-08-30
Adding sort button on properties and more


romilin:
Hello, 

I'd like to add a button on properties, Paths, Data Types and Responses. This will allow the user to choose if he wants to see these order alphabetically or order according to the source / priority. 

It would be great if we can also add a drag and drop option when ordered according to the source to let the user change the order .

What do you think about this feature ?

PR #2259, opened 2023-08-30
Add sort button

romilin:
Hello, 

This add a button on Properties, Path, Data types and Responses which let the user change the way these are sorted : either alphabetically or according to the source. 

It's the first part of the feature i would like to discuss in the issue #2258


EricWittmann (approved):

EricWittmann:
Seems reasonable, thanks. :)

PR #2274, opened 2023-09-11
Use the same Keycloak Realm env

dukex:
I'm using a custom Realm, following the documentation and in many places of the code, set up the `APICURIO_KC_REALM` env.

But some integrations, as the Github link account, still redirects the users to 'apicurio', the default realm, causing an error, because in my case we dont have the 'apicurio' realm.

Take a looking,  the env `APICURIO_KC_AUTH_REALM` still uses 'apicurio' as value, maybe an old code. 

Instead of the set-up the two env vars, my idea was to change the config to use the `APICURIO_KC_REALM`.

Maybe the props in this file an outdated too.


EricWittmann:
Are you saying there is just a disconnect between the documentation and reality?  And that this change is to change reality (the actual env var used by the app) to match the docs?

EricWittmann:
Closing without an answer to my question (please re-open if needed!).

Issue #1992, opened 2022-08-24
Unable to save api in source mode in v02.57-final


danylaberge:
Good day, this issue started with the release of 0.2.55-final and is still currently doing it. Steps leading to issue : 

1. Open and edit an API design
2. Go to source tab where you can see the full api source code
3. Do a change
4. Save changes
5. An error will occur
![image](https://user-images.githubusercontent.com/21954900/186427090-c180e11c-b4ec-45c2-aa77-a1e695b01ab0.png)

- Happens only in source of full api specifications
- doesn't happen in design mode
- doesn't happen in source mode when in individual path or components

We are running out of azure openshift and the application has been redeployed this morning with 0.2.57-final.

Thanks

EricWittmann:
Wow this is a strange one!  I've not seen anything like this before.  The Lost Connection warning would typically only happen if the websocket between the browser and editing server is broken.  

Some questions:

* Do you see any server logs in the apicurio-studio-ws component?
* How big are your API definitions?  If they are large can you try to reproduce this with a small (hello world) definition?

My best guess is some sort of web socket message size limit is being hit.

danylaberge:
I found the following in the ws logs

2022-08-25 20:05:36,377 WARN  [io.ver.cor.imp.BlockedThreadChecker] (vertx-blocked-thread-checker) Thread Thread[vert.x-eventloop-thread-1,5,main] has been blocked for 3034 ms, time limit is 2000 ms: io.vertx.core.VertxException: Thread blocked

2022-08-25 20:05:53,578 INFO  [io.api.hub.cor.cmd.OaiCommandExecutor] (vert.x-eventloop-thread-1) Applying 2 commands to a document.
|INFO| [ChangeDescriptionCommand] Executing.
|INFO| [AbstractReplaceNodeCommand] Executing.
2022-08-25 20:05:55,580 WARN  [io.net.cha.DefaultChannelPipeline] (vert.x-eventloop-thread-1) An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.: io.netty.handler.codec.http.websocketx.CorruptedWebSocketFrameException: Max frame length of 65536 has been exceeded.

- Yes, the API Specs are fairly large, mostly only schema definitions with a lot of external refs
- Doesn't seem to have an issue with small specs
- Reconnect doesn't work, changes will be lost when navigating out.

EricWittmann:
[...]

PR #2299, opened 2023-10-19
fix(mysql): increase websocket max frame size to handle full source editing

matthew-muscat:
- ensures users can edit the source of an api design and save it's contents over a websocket
- related to the fix introduced in https://github.com/Apicurio/apicurio-studio/commit/3f2a21360522ce4869f7f3d3c407b9eebd243fe8, but for mysql
- related to https://github.com/Apicurio/apicurio-studio/issues/1992

### DIFF
--- a/platforms/quarkus/ws/src/main/resources/application.properties
+++ b/platforms/quarkus/ws/src/main/resources/application.properties
@@ -34,5 +34,6 @@
 %mysql.quarkus.datasource.jdbc.max-size=8
 %mysql.quarkus.datasource.jdbc.min-size=2
 %mysql.quarkus.log.level=INFO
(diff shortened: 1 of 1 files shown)


matthew-muscat:
**Thought Bubble 💡** 

@EricWittmann 

Any reason why this configuration is not placed behind an environment variable, with a default value set if the env var is not set?

I'm assuming this can be interpreted on application startup (similar to the DB connection entries)

EricWittmann:
That property is interpreted at build time only:

https://quarkus.io/version/2.16/guides/all-config#quarkus-websockets-client_quarkus.websocket.max-frame-size

So an environment variable wouldn't help.

I think we should probably just set its value once rather than multiple times for each profile.  I'll merge this and we can think about that.

PR #2331, opened 2023-11-14
addition of a properties table for the reference data-types to be able to understand more clearly the data-type.

legas117:
reference

Adding a table showing the important informations of the referenced data-type. Would maybe be prettier with tabs to choose between the full file and the table view but I didn't manage to do that.


EricWittmann:
Interesting contribution, thank you!  I'll have a look soon.

EricWittmann (approved):

PR #2461, opened 2024-04-09
docs: initial Apicurio Studio documentation and install examples

jsenko:


app/src/main/java/io/apicurio/studio/auth/AuthorizedInterceptor.java
```diff
@@ -44,6 +45,7 @@ public class AuthorizedInterceptor {
     SecurityIdentity securityIdentity;
 
     @ConfigProperty(name = "quarkus.oidc.tenant-enabled")
+    @Info(category = "auth", description = "OIDC tenant enabled", availableSince = "1.0.0")
```
EricWittmann, line 48: Should probably read:  `OIDC authentication enabled`

EricWittmann: Hm - if we inject the same property in multiple places (as we do for this one) do we want to `@Info` describe in every place?

docs/generateAllConfigPartial.java
```diff
@@ -0,0 +1,310 @@
...
+import java.io.*;
+import java.util.*;
+import java.net.*;
+import java.nio.charset.*;
+import java.nio.file.*;
+import java.util.stream.Collectors;
+
+import static java.lang.System.*;
+
+public class generateAllConfigPartial {
```
EricWittmann, line 20: Is this something we could share across projects?

docs/modules/ROOT/assets/images/downstream/api-designer-design-options.png
EricWittmann, line 1: This will need to be updated to reflect the new upstream branding.

EricWittmann (dismissed):

docs/generateAllConfigPartial.java
jsenko: Probably, I'll look into adding it into common app components.

app/src/main/java/io/apicurio/studio/auth/AuthorizedInterceptor.java
jsenko: Do you know what is the difference between this property and `quarkus.oidc.enabled` ("OIDC enabled")?

docs/modules/ROOT/assets/images/downstream/api-designer-design-options.png
jsenko: Definitely, I didn't update the screenshots:)

PR #2462, opened 2024-04-11
Vendor extension support in OpenAPI via dynamic forms

EricWittmann:
It is now possible to configure Studio with a list of vendor extensions.  Each vendor extension has a name (e.g. `x-apicurio-address`), a JSON schema describing the format, a model with initial state, and an optional array of component types the extension applies to.  When configured, the OpenAPI editor will dynamically generate a form to create the value of the extension:

image

### DIFF
--- a/app/src/main/java/io/apicurio/studio/rest/v1/impl/SystemResourceImpl.java
+++ b/app/src/main/java/io/apicurio/studio/rest/v1/impl/SystemResourceImpl.java
@@ -1,13 +1,21 @@
 package io.apicurio.studio.rest.v1.impl;
 
+import java.io.InputStream;
+import java.net.URL;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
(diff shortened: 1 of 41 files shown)


EricWittmann:
Configuring the vendor extensions currently requires setting a URL to a JSON file that contains the details of the extensions.   The PR contains an example of such a file.

The URL to that file can be configured via an env variable - see the change to `UserInterfaceConfigProperties.java`.

@andreaTP @carlesarnal @jsenko What do you think about this as an approach to configure the extensions?  In the future I might change it to be a managed collection via the REST API, but that would need to be admin-only functionality and we don't currently have a concept of "admin" in Studio.

jsenko:
Docs need to be regenerated `docs/modules/ROOT/partials/ref-studio-all-configs.adoc`

EricWittmann:
> Docs need to be regenerated `docs/modules/ROOT/partials/ref-studio-all-configs.adoc`

Done.

PR #2576, opened 2024-05-30
Event feature

carlesarnal:
This PR starts the path for having events following the outbox pattern. This is just a wip with just a design-created event. Creating in draft for early feedback.

I'm planning on adding integration tests for this spinning up kafka and debezium proving that the expected events are sent and can be processed.


app/src/main/java/io/apicurio/studio/event/EventSender.java
```diff
@@ -0,0 +1,21 @@
...
+public class EventSender {
+
+    @Inject
+    StudioStorage storage;
+
+    public void onExportedEvent(@Observes OutboxEvent event) {
+        OutboxEvent createdEvent = storage.createEvent(event);
+
+        //Since log-based CDC is expected to be used, the events in the table can be immediately removed.
+        storage.deleteEvent(createdEvent.getId());
```
EricWittmann, line 19: Fascinating.  I didn't understand this point when you mentioned it in the call today.

carlesarnal: It really depends on how you configure the connector, but if it's configured to ignore the delete statements made to the outbox table, then you can do this, using a blasting 0 space in the disk.

carlesarnal: You can still send events with a "delete" semantic meaning, but that's still an insert to the outbox table.

app/src/main/resources-unfiltered/META-INF/storage/schema/upgrades/3/h2.upgrade.ddl
```diff
@@ -0,0 +1,8 @@
+-- *********************************************************************
+-- DDL for the Apicurio Studio - Database: H2
+-- Upgrades the DB schema from version 2 to version 3.
+-- *********************************************************************
+
+CREATE TABLE config (pname VARCHAR(255) NOT NULL, pvalue VARCHAR(1024), modifiedOn BIGINT NOT NULL);
+ALTER TABLE config ADD PRIMARY KEY (pname);
+CREATE INDEX IDX_config_1 ON config(modifiedOn);
```
EricWittmann, line 8: This doesn't look right.  :)

app/src/main/resources-unfiltered/META-INF/storage/schema/upgrades/3/postgresql.upgrade.ddl
```diff
@@ -0,0 +1,8 @@
+-- *********************************************************************
+-- DDL for the Apicurio Studio - Database: H2
+-- Upgrades the DB schema from version 2 to version 3.
+-- *********************************************************************
+
+CREATE TABLE config (pname VARCHAR(255) NOT NULL, pvalue VARCHAR(1024) NOT NULL, modifiedOn BIGINT NOT NULL);
+ALTER TABLE config ADD PRIMARY KEY (pname);
+CREATE INDEX IDX_config_1 ON config(modifiedOn);
```
EricWittmann, line 8: Same

spi/src/main/java/io/apicurio/studio/spi/storage/model/OutboxEvent.java
```diff
@@ -0,0 +1,11 @@
+package io.apicurio.studio.spi.storage.model;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public interface OutboxEvent {
+    String getId();
+    String getAggregateId();
+    String getAggregateType();
```
EricWittmann, line 8: What are these?  The `aggregateId` seems to be the native ID of the "thing" the event represents?  So in this case the `designId`.  Yes?  What's `aggregateType` ?

app/src/main/resources-unfiltered/META-INF/storage/schema/upgrades/3/h2.upgrade.ddl
carlesarnal: Ouch, copy-paste error. I have added the table to the main script and the upgrade is not being tested.

spi/src/main/java/io/apicurio/studio/spi/storage/model/OutboxEvent.java
carlesarnal: AggregateId is the id of the design, yes. `aggregateType` is the "type" of the outbox event, it's the value that eventually allows the connector to route the event to the appropriate topic. In the spirit of flexibility of the solution, I think it's worth having it even if we decide to go with the single-topic approach.

app/src/main/java/io/apicurio/studio/event/EventSender.java
EricWittmann: Yeah understood.  It's very interesting. Does this work for all DBs supported by DBZ?

spi/src/main/java/io/apicurio/studio/spi/storage/model/OutboxEvent.java
EricWittmann: What are some examples of  the value of `aggregateType`?  And what is the semantic meaning of `aggregate` in these variables?  Not sure I fully understand yet...

app/src/main/java/io/apicurio/studio/event/EventSender.java
carlesarnal: Not sure if it works for all, I think so, since it's the transformation that does this (ignoring the delete event). I can check with the DBZ folks.

spi/src/main/java/io/apicurio/studio/spi/storage/model/OutboxEvent.java
carlesarnal: Aggregate is the source table of the event. In Studio, in this case, we're only tracking the designs, so an example aggregate would be the design. It may be users if we decide to send events for those. Any entity where you're tracking events. The aggregateType is something that defines it, in this case, I'm using "DESIGN", which is usually used to separate the different events into separate topics. But, since we don't want that, we can probably use something like "studio".

carlesarnal:
@EricWittmann any idea on how you want to proceed with this? Our events thing won't look like it very likely, so maybe just close this PR?

EricWittmann:
Yes I think let's go ahead and close this PR for now. 

Based on current thinking this will probably not be needed in studio at least not for now.

PR #2888, opened 2024-10-16
Add explore list with groups, artifact view and versions and branches lists.

carlesarnal:


EricWittmann (approved):

EricWittmann:
No need for a Version Details page.  Also we probably want to streamline the other pages.  We don't need everything we have in Registry. We just want to navigate the hiearchy.  So just basic metadata + list of versions is all we need for an Artifact (for example).  e.g. no need to show the rules

PR #2966, opened 2024-11-27
fix(asyncapi): broken schema inheritance

ben-lc:
For asyncapi docs, defining schema with oneOf, allOf or anyOf inheritance isn't working cause required apicurio-data-models commands are missing and `definition-form.component.ts` only supports openapi docs.

This PR upgrades apicurio-data-models to 1.1.30 which brings required commands and adds required modifications to support it.

Marked as draft until  apicurio-data-models 1.1.30 is released.

### DIFF
--- a/front-end/studio/package.json
+++ b/front-end/studio/package.json
@@ -18,7 +18,7 @@
   "private": false,
   "peerDependencies": {
     "@patternfly/patternfly": "1.0.250",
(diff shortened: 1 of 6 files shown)


EricWittmann:
Releasing 1.1.30 now.

EricWittmann (approved):

